Issue #249, opened 2024-02-21
extracting part of a cell, network etc.


jnsbck:
Hi,

I have a morphology 

```python

cell = jx.read_swc(f"morphology.swc", nseg=4, max_branch_len=300.0, min_radius=5.0)
cell.insert(HH())
```

but I want to simulate only a specific branch. I don't care about anything else that is attached. Since `cell.branch(0)` returns an instance of `BranchView`, the following does not work.

```python
branch = cell.branch(0)

branch.comp(0).stimulate(current)
branch.record()

voltage = jx.integrate(branch)
```

Is this somehow possible, since I was not able to find an easy way to do this? If not it would be awesome to have this functionality.

i.e. by adding a `.copy()` method to the `BranchView` object, that returns a new instance of `Branch`

```python
branch = cell.branch(0).copy()

branch..comp(0).stimulate(current)
branch.record()

voltage = jx.integrate(branch)
```
Same would be applicable to `Cell` and `Comp`. Would be happy to provide a PR if this functionality is otherwise non-trivially obtained.

Thanks for your input on this. :)

michaeldeistler:
Related to #354

jnsbck:
This can be done in a limited fashion with #447 and should be improved down the line.

jnsbck:
the following works if #487 is merged:

```python
def infer_module_type_from_inds(idxs: pd.DataFrame) -> str:
    nuniques = idxs[["cell_index", "branch_index", "comp_index"]].nunique()
    nuniques.index = ["cell", "branch", "compartment"]
    nuniques = pd.concat([pd.Series({"network": 1}), nuniques])
    return_type = nuniques.loc[nuniques == 1].index[-1]
    return return_type


def build_module_scaffold(
    idxs: pd.DataFrame,
    return_type: Optional[str] = None,
    parent_branches: Optional[List[np.ndarray]] = None,
) -> Union[jx.Network, jx.Cell, jx.Branch, jx.Compartment]:
    """Builds a skeleton module from a DataFrame of indices.
    This is useful for instantiating a module that can be filled with data later.
    Args:
        idxs: DataFrame containing cell_index, branch_index, comp_index, i.e.
            Module.nodes or View.view.
        return_type: Type of module to return. If None, the type is inferred from the
            number of unique values in the indices. I.e. only 1 unique cell_index
                and 1 unique branch_index -> return_type = "jx.Branch".
    Returns:
[...]

PR #351, opened 2024-04-24
Make `View` behave more like `Module`

jnsbck:
I have added a bunch of attrs that can be accessed in `Module` as properties of `View`. Calling `View.attr` behaves similar to `Module.attr`, but only considers whatever parts of the network, cell, etc. are currently in view. This could potentially allow to run `integrate` on `View`, i.e. if one just wants to look at the `soma` of a cell or a single cell in a network. This would also make it easier to add a `to_module` util, that can take a `View` and turn it into a standalone module, which was suggested in #249.

Also added a `groups` attr, that just lists group names.

Open questions and todos are:
- [ ] Add tests to check if View and Module work the same way!
- [ ] Check if integration can be run on View
- [ ] Decide how to deal with disconnected views. Potentially add `_ensure_view_is_connected`?
- [x] Remove xyzr if #320 gets merged
- [ ] How to deal with nseg if only parts of branch are viewed or when viewing single 2 comps or 2 branch sections in 2 cells
- [ ] rid `View.nodes` of unused parameters
- [ ] handle trainable_params
- [ ] Are all relevant / important properties accessible? Do some need to be added on `CellView` etc. level?
- [ ] remove `dev.ipynb`

In the long term, it would be nice to make `View` behave more like `Module` without all of these properties essentially just indexing into `View.pointer` in some way. Imo most methods should be part of module if possible and only passed through to view, with view only showing what's in scope. While a method in module acts on the entire thing, in view it only acts on whatever is in view. This could be done by having a `self.viewed_indexes` / `self.in_scope` attr in `Module`, that decides what is `set`, where things are `insert`ed etc. This would get rid of a lot of complexity and duplicate methods. For example would this make many hidden methods, like `_record` in `Module` obsolete. Additionally `view` can just be a property that returns `View.nodes`.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1386,6 +1386,141 @@ class View:
         self.pointer.synapse_state_names += list(synapse_type.synapse_states.keys())
         self.pointer.synapses.append(synapse_type)
 
+    def _viewed_indices(self, level: str = "comp"):
+        """Unique compartment, branch, and cell indicies currently in view."""
+        return np.unique(self.view[f"{level}_index"])
+    
+    # make View behave like a Module
+    # TODO: Add tests to check if View and Module work the same way!
+    # comp = Comp(); viewed_comp = net.comp(0)
+    # for attr in ["attr1", ...]:; assert comp.attr == viewed_comp.attr
+    # repeat for different views of comps, branches, cells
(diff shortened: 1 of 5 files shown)


jnsbck:
Much better implementation in #447

Issue #336, opened 2024-04-12
API for setting individual synaptic parameters


kyralianaka:
Currently @Matthijspals and I index into network.edges to change individual synaptic parameters like gS and k_minus, but it would be cool to make this part of the API so that one could just specify the pre and post cell (and I guess the synapse type in case there are multiple synapse types between the pre and post cell), the parameter and to what it should be set to. 

One can currently do this for a certain parameter of *all* the synapses of that type, and maybe one can give array input also to adjust things individually (not sure, will have to test this more thoroughly), but it might be nice to expand on this.

michaeldeistler:
This is enabled by #447

PR #360, opened 2024-05-03
Method for getting synapse indices

### LINKED ISSUE #264: enable synapse param sharing API
E.g.
```python
synapse_indices  = get_indices(net, edge_prop_names)
net.synapses(synapse_indices).add_to_group(group_name)
net.group_name.make_trainable("gS")
```

IMG_7106

kyralianaka:
I wrote a method in network.py for getting the indices of synapses in the SynapseView given their edge properties. I'm happy to put it somewhere else though if someone prefers. This addresses two git issues, #336 and #264, but it doesn't fully resolve #264, rather just the first step. 

I do find it a bit odd that there is not one SynapseView per network with all of the synapses, and rather that one indexes into the different synapse types separately. Especially since the parameters are even explicitly prefixed by the synapse type. I figured the full ramifications of changing to a unified synapse view would require a longer conversation though, and it doesn't seem particularly necessary, just nicer to look at. 

I am also happy to change the argument type, because I know that passing a dictionary to a method is very cumbersome. I just wanted to make it flexible as described in #264 :)

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1093,6 +1093,42 @@ class Module(ABC):
 
         self._update_nodes_with_xyz()
 
+    def get_synapse_indices(
+        self,
+        pre_cells: Union[int, List],
+        post_cells: Union[int, List],
+        synapse_type: str,
+    ) -> List[int]:
+        """Get the synapse indices in the SynapseView between the given pre and post synaptic cells.
+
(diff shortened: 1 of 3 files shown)


kyralianaka:
Synapse indexing is about to get overhauled with the removal of the synapse view, and something like this will get moved to a utilities file as I understand it

PR #380, opened 2024-07-15
add: add clone tridiax and pytest actions.

jnsbck:
We can either push `tridiax` to pypi or just install it from source. The latter would requite generating a personal access token in the org, which unfortunately I don't have permission to do. See here for how to do it. [Clone Github Repository Action](https://github.com/marketplace/actions/clone-github-repo-action).


jnsbck:
Will close this since tridiax on pypi now and tests are part of ci

PR #395, opened 2024-08-05
allow to not update nodes after move for speed

michaeldeistler:
For large networks `update_nodes_with_xyzr` is so slow that I had to revert to old commits to visualize my networks.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -10,7 +10,7 @@ import jax.numpy as jnp
 import networkx as nx
 import numpy as np
 import pandas as pd
-from jax import vmap
+from jax import jit, vmap
 from jax.lax import ScatterDimensionNumbers, scatter_add
 from matplotlib.axes import Axes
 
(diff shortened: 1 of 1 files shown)


jnsbck:
Wow, changing the `jnp` to `np` made this almost another order of mag faster! Will make a PR. Would be interested to see if this makes it feasable to run on your network.

```python
v_interp = jit(vmap(jnp.interp, in_axes=(None, None, 1)))

def update_nodes_with_xyz(self):
    """Add xyz coordinates to nodes."""
    num_branches = len(self.xyzr)
    x = np.linspace(0.5 / self.nseg, (num_branches*1 - 0.5 / self.nseg), num_branches*self.nseg)
    x += np.arange(num_branches).repeat(net.nseg) # add offset to prevent branch loc overlap
    xp = np.hstack([np.linspace(0, 1, x.shape[0])+2*i for i,x in enumerate(self.xyzr)])
    xyz = v_interp(x, xp, np.vstack(self.xyzr)[:,:3])
    idcs = self.nodes["comp_index"]
    self.nodes.loc[idcs, ["x", "y", "z"]] = xyz.T
    return xyz.T
```

PR #401, opened 2024-08-15
Add test for license header and add missing headers

jnsbck:
noticed that not all files had a license header so I added this to the CI.


michaeldeistler (approved):
LGTM, thanks!

PR #411, opened 2024-08-27
fix  equidistant interpolation in xyz computation

jnsbck:
the current implementation of `compute_xyz` does not correctly compute the locations of the xyz coordinates, according to their position along the traced path to be interpolated, but just sets `xp=linspace(0,1,len(x))`.

I.e. currently there is a missmatch between the xyz coordinates (fp) and the branchlens/pathlens (xp):

xyz1-----xyz2------xyz3-----------------xyz4--xyz5
loc1--------loc2--------loc3--------loc4--------loc5

rather than taking into account the correct pathlens

xyz1-----xyz2------xyz3-----------------xyz4--xyz5
loc1-----loc2------loc3-----------------loc4--loc5

### DIFF
--- a/jaxley/utils/cell_utils.py
+++ b/jaxley/utils/cell_utils.py
@@ -295,8 +295,8 @@ def interpolate_xyz(loc: float, coords: np.ndarray):
         Interpolated xyz coordinate at `loc`, shape `(3,).
     """
     dl = np.sqrt(np.sum(np.diff(coords[:, :3], axis=0) ** 2, axis=1))
-    pathlens = np.insert(np.cumsum(dl), 0, 0) # cummulative length of sections
-    norm_pathlens = pathlens/pathlens[-1]  # path lengths normalized to [0,1]
+    pathlens = np.insert(np.cumsum(dl), 0, 0)  # cummulative length of sections
+    norm_pathlens = pathlens / pathlens[-1]  # path lengths normalized to [0,1]
 
     return v_interp(loc, norm_pathlens, coords[:, :3])
 


jaxley/utils/cell_utils.py
```diff
@@ -291,9 +294,9 @@ def interpolate_xyz(loc: float, coords: np.ndarray):
     Return:
         Interpolated xyz coordinate at `loc`, shape `(3,).
     """
-    return vmap(lambda x: jnp.interp(loc, jnp.linspace(0, 1, len(x)), x), in_axes=(1,))(
-        coords[:, :3]
-    )
+    lens = np.cumsum(np.sqrt(np.sum(np.diff(coords[:, :3], axis=0) ** 2, axis=1)))
```
michaeldeistler, line 297: why is the variable name `lens`?

michaeldeistler (commented):
Could you provide some more context? What does this fix? Also I really don't understand what is happening in your code, please add comments

jaxley/utils/cell_utils.py
```diff
@@ -291,9 +294,11 @@ def interpolate_xyz(loc: float, coords: np.ndarray):
     Return:
         Interpolated xyz coordinate at `loc`, shape `(3,).
     """
-    return vmap(lambda x: jnp.interp(loc, jnp.linspace(0, 1, len(x)), x), in_axes=(1,))(
```
jnsbck, line 294: `jnp.interp` takes `x`, `xp` and `fp`. The crux is that in the current implementation `xp=jnp.linspace(0, 1, len(x)), x)`. This would mean every `fp` i.e. the xyz coords along which we want to interpolate are equidistant (i.e. the pathlengths between them are the same), which they usually are not!

jaxley/utils/cell_utils.py
```diff
@@ -291,9 +294,11 @@ def interpolate_xyz(loc: float, coords: np.ndarray):
...
-        coords[:, :3]
-    )
+    dl = np.sqrt(np.sum(np.diff(coords[:, :3], axis=0) ** 2, axis=1))
+    pathlens = np.insert(np.cumsum(dl), 0, 0) # cummulative length of sections
+    norm_pathlens = pathlens/pathlens[-1]  # path lengths normalized to [0,1]
```
jnsbck, line 299: We instead want to interpolate along the traced out path, according to the pathlengths between the coords. Hence `xp` should correspond to the locs at each xyz coord.

jnsbck:
Sorry for the brevity. Does this make more sense now? @michaeldeistler

jaxley/utils/cell_utils.py
michaeldeistler: Ohhh I see! Thanks!

michaeldeistler (approved):
Thanks a ton!

PR #418, opened 2024-09-13
Allow different nseg per branch and add `jax.sparse` solver

michaeldeistler:
# New features

### Different nseg for every branch

This PR enables that every branch can have a different number of compartments.
```python
comp = jx.Compartment()
branch1 = jx.Branch(comp, nseg=1)
branch2 = jx.Branch(comp, nseg=2)
branch3 = jx.Branch(comp, nseg=3)
cell = jx.Cell([branch1, branch2, branch3], parents=[-1, 0, 0])
cell.insert(HH())

cell.branch(1).comp(1).stimulate(jx.step_current(0.5, 1.0, 0.1, 0.025, 5.0))
cell.branch(0).comp(0).record()

voltages = jx.integrate(cell, delta_t=dt, voltage_solver="jax.sparse")
```

This is achieved by:
1) Building infrastructure to deal with a different number of compartments and
2) relying on `jax.experimental.sparse.spsolve` as a generic solver.

### Generic sparse voltage solver

With this PR, we are now supporting `jx.integrate(..., voltage_solver="jax.sparse")`. This uses the generic [sparse solver of JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.sparse.linalg.spsolve.html) for solving the voltage equations. It has the advantage that it has zero compile time and can more easily deal with different nseg, but the disadvantage that it does not support `vmap`. Unfortunately, the current voltage solvers (`jaxley.thomas` and `jaxley.stone`) need a different format for the sparse voltage solver than the `jax.sparse` solver. Therefore, we have to support both. This is achieved by:

1) Splitting `Module.init_morph()` into `Module._init_morph_jax_spsolve()` and `Module._init_morph_custom_spsolve()` (the latter is the one we already had).
2) Splitting `Module.init_conds()` into `Module._init_conds_jax_spsolve()` and `Module._init_conds_custom_spsolve()` (the latter is the one we already had).
3) Splitting `step_voltage_implicit()` (which lives in `jaxley.solver_voltage.py`) into `step_voltage_implicit_with_jax_spsolve()` and `step_voltage_implicit_with_custom_spsolve()`.

# Follow-up work

### Required future work before releasing
- Currently, `nseg` cannot be set after initialization. We should allow something like `cell.branch.set("nseg", 4)`.
- In a future PR, I will also allow using `jaxley.thomas` and `jaxley.stone` with different `nseg` per branch by masking out unused dimensions. This is done in #426 

### Optional follow-up work (probably after the next release)
- Raise an explicit error when `vmap` is used around the sparse jax solver.
- Consider making `jax.sparse` the default solver because of its low compile time.
- Benchmark the solvers and write a short tutorial about it.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -256,8 +256,31 @@ class Module(ABC):
 
         return printable_nodes
 
+    def init_morph(self):
+        """Initialize the morphology such that it can be processed by the solvers."""
+        self._init_morph_custom_spsolve()
+        self._init_morph_jax_spsolve()
+        self.initialized_morph = True
+
     @abstractmethod
-    def init_conds_jax_spsolve(self, params: Dict):
+    def _init_morph_jax_spsolve(self):
+        """Initialize the morphology for the JAX sparse solver."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def _init_morph_custom_spsolve(self):
+        """Initialize the morphology for the custom Jaxley solver."""
+        raise NotImplementedError
+
+    def init_conds(self, params: Dict, voltage_solver: str):
+        """Given radius, length, r_a, compute the axial coupling conductances."""
+        if voltage_solver.startswith("jaxley"):
+            return self._init_conds_custom_spsolve(params)
+        else:
+            return self._init_conds_jax_spsolve(params)
+
+    @abstractmethod
+    def _init_conds_jax_spsolve(self, params: Dict):
         """Initialize coupling conductances.
 
         Args:
@@ -267,7 +290,7 @@ class Module(ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def init_conds_custom_spsolve(self, params: Dict):
+    def _init_conds_custom_spsolve(self, params: Dict):
         """Initialize coupling conductances.
 
         Args:
@@ -554,10 +577,7 @@ class Module(ABC):
                 params[key] = params[key].at[inds].set(set_param[:, None])
 
         # Compute conductance params and append them.
-        if voltage_solver.startswith("jaxley"):
-            cond_params = self.init_conds_custom_spsolve(params)
-        else:
-            cond_params = self.init_conds_jax_spsolve(params)
+        cond_params = self.init_conds(params=params, voltage_solver=voltage_solver)
         for key in cond_params:
             params[key] = cond_params[key]
 
@@ -620,9 +640,7 @@ class Module(ABC):
 
     def initialize(self):
         """Initialize the module."""
-        self.init_morph_custom_spsolve()
-        self.init_morph_jax_spsolve()
-        self.initialized_morph = True
+        self.init_morph()
         return self
 
     def init_states(self, delta_t: float = 0.025):
@@ -880,9 +898,6 @@ class Module(ABC):
     def init_syns(self):
         self.initialized_syns = True
 
-    def init_morph(self):
-        self.initialized_morph = True
-
     def step(
         self,
         u: Dict[str, jnp.ndarray],
@@ -954,12 +969,12 @@ class Module(ABC):
                 "voltage_terms": (v_terms + syn_v_terms) / cm,
                 "constant_terms": (const_terms + i_ext + syn_const_terms) / cm,
                 "axial_conductances": params["axial_conductances"],
-                "data_inds": self.data_inds,
-                "indices": self.indices,
-                "indptr": self.indptr,
-                "sinks": np.asarray(self.comp_edges["sink"].to_list()),
-                "n_nodes": self.n_nodes,
-                "internal_node_inds": self.internal_node_inds,
+                "data_inds": self._data_inds,
+                "indices": self._indices_jax_spsolve,
+                "indptr": self._indptr_jax_spsolve,
+                "sinks": np.asarray(self._comp_edges["sink"].to_list()),
+                "n_nodes": self._n_nodes,
+                "internal_node_inds": self._internal_node_inds,
             }
             # Only for `bwd_euler` and `cranck-nicolson`.
             step_voltage_implicit = step_voltage_implicit_with_jax_spsolve
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -15,19 +15,24 @@ from jax.lax import ScatterDimensionNumbers, scatter_add
 from matplotlib.axes import Axes
 
 from jaxley.channels import Channel
-from jaxley.solver_voltage import step_voltage_explicit, step_voltage_implicit
+from jaxley.solver_voltage import (
+    step_voltage_explicit,
+    step_voltage_implicit_with_custom_spsolve,
+    step_voltage_implicit_with_jax_spsolve,
+)
 from jaxley.synapses import Synapse
 from jaxley.utils.cell_utils import (
     _compute_index_of_child,
     _compute_num_children,
     compute_levels,
     convert_point_process_to_distributed,
+    convert_to_csc,
     interpolate_xyz,
     loc_of_index,
     query_channel_states_and_params,
     v_interp,
 )
-from jaxley.utils.debug_solver import compute_morphology_indices, convert_to_csc
+from jaxley.utils.debug_solver import compute_morphology_indices
 from jaxley.utils.misc_utils import childview, concat_and_ignore_empty
 from jaxley.utils.plot_utils import plot_morph
 
@@ -252,7 +257,17 @@ class Module(ABC):
         return printable_nodes
 
     @abstractmethod
-    def init_conds(self, params: Dict):
+    def init_conds_jax_spsolve(self, params: Dict):
+        """Initialize coupling conductances.
+
+        Args:
+            params: Conductances and morphology parameters, not yet including
+                coupling conductances.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def init_conds_custom_spsolve(self, params: Dict):
         """Initialize coupling conductances.
 
         Args:
@@ -482,7 +497,9 @@ class Module(ABC):
         """
         return self.trainable_params
 
-    def get_all_parameters(self, pstate: List[Dict]) -> Dict[str, jnp.ndarray]:
+    def get_all_parameters(
+        self, pstate: List[Dict], voltage_solver: str
+    ) -> Dict[str, jnp.ndarray]:
         """Return all parameters (and coupling conductances) needed to simulate.
 
         Runs `init_conds()` and return every parameter that is needed to solve the ODE.
@@ -502,7 +519,13 @@ class Module(ABC):
 
         Args:
             pstate: The state of the trainable parameters. pstate takes the form
-                [{"key": "gNa", "indices": jnp.array([0, 1, 2]), "val": jnp.array([0.1, 0.2, 0.3])}, ...].
+                [{
+                    "key": "gNa", "indices": jnp.array([0, 1, 2]),
+                    "val": jnp.array([0.1, 0.2, 0.3])
+                }, ...].
+            voltage_solver: The voltage solver that is used. Since `jax.sparse` and
+                `jaxley.xyz` require different formats of the axial conductances, this
+                function will default to different building methods.
 
         Returns:
             A dictionary of all module parameters.
@@ -531,7 +554,10 @@ class Module(ABC):
                 params[key] = params[key].at[inds].set(set_param[:, None])
 
         # Compute conductance params and append them.
-        cond_params = self.init_conds(params)
+        if voltage_solver.startswith("jaxley"):
+            cond_params = self.init_conds_custom_spsolve(params)
+        else:
+            cond_params = self.init_conds_jax_spsolve(params)
         for key in cond_params:
             params[key] = cond_params[key]
 
@@ -594,7 +620,9 @@ class Module(ABC):
 
     def initialize(self):
         """Initialize the module."""
-        self.init_morph()
+        self.init_morph_custom_spsolve()
+        self.init_morph_jax_spsolve()
+        self.initialized_morph = True
         return self
 
     def init_states(self, delta_t: float = 0.025):
@@ -611,7 +639,9 @@ class Module(ABC):
 
         # We do not use any `pstate` for initializing. In principle, we could change
         # that by allowing an input `params` and `pstate` to this function.
-        params = self.get_all_parameters([])
+        # `voltage_solver` could also be `jax.sparse` here, because both of them
+        # build the channel parameters in the same way.
+        params = self.get_all_parameters([], voltage_solver="jaxley.thomas")
 
         for channel in self.channels:
             name = channel._name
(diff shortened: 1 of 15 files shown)


jaxley/modules/cell.py
```diff
@@ -251,8 +331,48 @@ def init_conds(self, params: Dict) -> Dict[str, jnp.ndarray]:
...
+            params["radius"][sink_comp_inds],
+            params["axial_resistivity"][source_comp_inds],
+            params["axial_resistivity"][sink_comp_inds],
+            params["length"][source_comp_inds],
+            params["length"][sink_comp_inds],
+        )
+
+        # `branchpoint-to-compartment` conductances.
+        condition = self.comp_edges["type"].to_numpy() == 1
+        sink_comp_inds = np.asarray(self.comp_edges[condition]["sink"].to_list())
```
jnsbck, line 352: is there are reason for doing `np.asarray(x.to_list())` rather than just `to_numpy()`

jnsbck: could potentially be done with `.groupby("type")` as well but not sure.

jnsbck (changes requested):
Have not managed to go through everything yet. Will look at the rest tmr. Have so far mostly made minor, localized comments. Wanna take a look at the overall structure of this when I am done reading through everything. From what I have so far seen the main changes seem to be:
1. split inits of morph / conds into seperate methods to conform with solver inputs
2. add comp_edges (incl branchpoints) and comp_edges_to_indices

Overall there is a lot going on here. Superficially at least there is a ton of vmaps over params, pd.concats and list comprehensions that I makes me think whether there is not some way to keep everything organised neatly in a single jax_sp_solver dataframe or something similar (likewise for the custom solver potentially). Need to do some more reading tmr though. Hoping that you can make some headway with this initial feedback though. 
Lemme know if you want feedback on sth specific or any further explanations.

jaxley/modules/cell.py
```diff
@@ -76,9 +77,9 @@ def __init__(
         parents = [-1] if parents is None else parents
 
         if isinstance(branches, Branch):
-            branch_list = [branches for _ in range(len(parents))]
+            self.branch_list = [branches for _ in range(len(parents))]
         else:
-            branch_list = branches
+            self.branch_list = branches
```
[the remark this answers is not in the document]
michaeldeistler: Yes, `.init_conds()` is called in `.initialize()`, which itself gets called at the end of `__init__()`.

jaxley/modules/cell.py
```diff
@@ -122,19 +127,15 @@ def __init__(
         )
 
         # For morphology indexing.
-        par_inds = self.branch_edges["parent_branch_index"].to_numpy()
-        self.child_inds = self.branch_edges["child_branch_index"].to_numpy()
+        par_inds = self.comb_parents[1:]
+        self.child_inds = np.arange(1, self.total_nbranches)
```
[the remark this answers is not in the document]
michaeldeistler: haha me neither, have to have a look.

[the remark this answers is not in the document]
michaeldeistler: Alright, I remember, I had thought I would delete `.branch_edges`, so I changed this. Then decided to keep `.branch_edges`, but forgot to change it back. Doing that now.

jaxley/modules/cell.py
```diff
@@ -183,9 +185,87 @@ def init_morph(self):
...
+                "type": 2,
+            }
+        )
+        branchpoint_to_parent_edges = pd.DataFrame().from_dict(
+            {
+                "source": np.arange(len(self.par_inds)) + self.cumsum_nseg[-1],
+                "sink": self.cumsum_nseg[self.par_inds + 1] - 1,
+                "type": 2,
+            }
+        )
```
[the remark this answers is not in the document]
michaeldeistler: True!

jaxley/modules/base.py
```diff
@@ -918,43 +948,66 @@ def step(
         # Voltage steps.
         cm = params["capacitance"]  # Abbreviation.
 
-        solver_kwargs = {
-            "voltages": voltages,
-            "voltage_terms": (v_terms + syn_v_terms) / cm,
-            "constant_terms": (const_terms + i_ext + syn_const_terms) / cm,
```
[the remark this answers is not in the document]
michaeldeistler: will leave this for the future as it might change in the follow-up PR

jaxley/modules/base.py
```diff
@@ -918,43 +948,66 @@ def step(
...
+                "voltages": voltages,
+                "voltage_terms": (v_terms + syn_v_terms) / cm,
+                "constant_terms": (const_terms + i_ext + syn_const_terms) / cm,
+                "coupling_conds_upper": params["branch_uppers"],
+                "coupling_conds_lower": params["branch_lowers"],
+                "summed_coupling_conds": params["branch_diags"],
+                "branchpoint_conds_children": params["branchpoint_conds_children"],
+                "branchpoint_conds_parents": params["branchpoint_conds_parents"],
+                "branchpoint_weights_children": params["branchpoint_weights_children"],
+                "branchpoint_weights_parents": params["branchpoint_weights_parents"],
```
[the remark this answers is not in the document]
michaeldeistler: Will also leave this for the future

jaxley/modules/cell.py
michaeldeistler: Yes, because somehow `to_numpy` returns a list of Arrays here...I still have no idea why, and I was too lazy to dive into it in detail.

michaeldeistler:
This was great feedback, thanks a lot! All addressed now.

jaxley/solver_voltage.py
```diff
@@ -62,10 +63,10 @@ def step_voltage_explicit(
         debug_states,
     )
     new_voltates = voltages + delta_t * update
-    return new_voltates
+    return new_voltates.ravel(order="C")
 
 
-def step_voltage_implicit(
+def step_voltage_implicit_with_custom_spsolve(
```
jnsbck, line 69: if you want you could also replace custom with jaxley ;)

jnsbck: jaxley_spsolve vs jax_spsolve

jaxley/solver_voltage.py
```diff
@@ -196,8 +197,50 @@ def step_voltage_implicit(
...
+
+    # if-case needed because `.at` does not allow empty inputs, but the input is
+    # empty for compartments.
+    if len(sinks) > 0:
+        diagonal_values = diagonal_values.at[sinks].add(delta_t * axial_conductances)
+
+    diagonal_values = diagonal_values.at[internal_node_inds].add(
+        delta_t * voltage_terms
+    )
+    diagonal_values = diagonal_values.at[internal_node_inds].add(1.0)
```
jnsbck, line 227: does `.add.add` work?

jaxley/solver_voltage.py
```diff
@@ -196,8 +197,50 @@ def step_voltage_implicit(
...
+
+    # Build off-diagonals.
+    axial_conductances = -delta_t * axial_conductances
```
jnsbck, line 230: could be computed earlier and used in L222.

jaxley/utils/cell_utils.py
```diff
@@ -411,3 +407,134 @@ def query_channel_states_and_params(d, keys, idcs):
...
+    np.add.at(indptr, col_ind + 1, 1)
+    np.cumsum(indptr, out=indptr)
+
+    # Step 3: The row indices are already sorted
+    indices = row_ind
+
+    return data_inds, indices, indptr
+
+
+def comp_edges_to_indices(
```
jnsbck, line 445: does it make sense to keep `convert_to_csc` and `comp_edges_to_indices` in cell_utils? Would move to misc_utils potentially or add solver_utils.

jaxley/utils/cell_utils.py
```diff
@@ -411,3 +407,134 @@ def query_channel_states_and_params(d, keys, idcs):
...
+    return n_nodes, data_inds, indices, indptr
+
+
+def compute_axial_conductances(
+    comp_edges: pd.DataFrame, params: Dict[str, jnp.ndarray]
+) -> jnp.ndarray:
+    """Given `comp_edges`, radius, length, r_a, compute the axial conductances."""
+    # `Compartment-to-compartment` (c2c) conductances.
+    condition = comp_edges["type"].to_numpy() == 0
+    source_comp_inds = np.asarray(comp_edges[condition]["source"].to_list())
```
jnsbck, line 487: maybe doing comp_edges.loc[condition,source] fixes the array/list issue

jnsbck: you could also just do `source_sink_inds = comp_edges.loc[condition, ["source", "sink"]]` and then do
`*params["radius"][source_sink_inds]` in 491 (I think this should work. Not sure though).

jnsbck, line 487: could also make this a ` at_inds = lambda: type, key: comp_edges.loc[comp_edges["type"].to_numpy() == type, key]`

jaxley/utils/cell_utils.py
```diff
@@ -411,3 +407,134 @@ def query_channel_states_and_params(d, keys, idcs):
...
+    sink_comp_inds = np.asarray(comp_edges[condition]["sink"].to_list())
+
+    conds_c2c = vmap(compute_coupling_cond, in_axes=(0, 0, 0, 0, 0, 0))(
```
jnsbck, line 490:
```
param_keys = ['radius', 'axial_resistivity', 'length']
conds_c2c = vmap(compute_coupling_cond)(
    *[params[key][inds] for key in param_keys for inds in (sink_comp_inds, source_comp_inds)]
)
```
not sure if necessary, but this could also work

jaxley/utils/cell_utils.py
```diff
@@ -411,3 +407,134 @@ def query_channel_states_and_params(d, keys, idcs):
...
+        params["length"][source_comp_inds],
+    )
+
+    # `branchpoint-to-compartment` (bp2c) conductances.
+    condition = comp_edges["type"].to_numpy() == 1
+    sink_comp_inds = np.asarray(comp_edges[condition]["sink"].to_list())
+
+    if len(sink_comp_inds) > 0:
+        conds_bp2c = vmap(compute_coupling_cond_branchpoint, in_axes=(0, 0, 0))(
+            params["radius"][sink_comp_inds],
```
jnsbck, line 505: ... param keys could be reused later as well

jaxley/utils/cell_utils.py
```diff
@@ -411,3 +407,134 @@ def query_channel_states_and_params(d, keys, idcs):
...
+    else:
+        conds_bp2c = jnp.asarray([])
+
+    # `compartment-to-branchpoint` (c2bp) conductances.
+    condition = comp_edges["type"].to_numpy() == 2
+    source_comp_inds = np.asarray(comp_edges[condition]["source"].to_list())
+
+    if len(source_comp_inds) > 0:
+        conds_c2bp = vmap(compute_impact_on_node, in_axes=(0, 0, 0))(
+            params["radius"][source_comp_inds],
```
jnsbck, line 518: ...and here

jaxley/utils/cell_utils.py
jnsbck, line 490: rename `coupling_conds_c2c`?

tests/jaxley_identical/test_basic_modules.py
```diff
@@ -131,7 +135,42 @@ def test_cell():
     assert max_error <= tolerance, f"Error is {max_error} > {tolerance}"
 
 
-def test_net():
+def test_cell_unequal_compartment_number():
```
jnsbck, line 138: Maybe also add `jaxley.stone` here and test that it correctly raises.

tests/jaxley_vs_neuron/test_cell.py
```diff
@@ -130,3 +130,116 @@ def integrate():
 
     voltages = np.asarray([list(voltage1), list(voltage2), list(voltage3)])
     return voltages
+
+
+def test_similarity_unequal_number_of_compartments():
```
jnsbck, line 135: Cool!

tests/test_api_equivalence.py
```diff
@@ -186,3 +186,46 @@ def test_api_equivalence_continued_simulation():
 
     v2 = jnp.concatenate([v21, v22[:, 1:]], axis=1)
     assert np.max(np.abs(v1 - v2)) < 1e-8
+
+
+def test_api_equivalence_network_matches_cell():
```
jnsbck, line 191: isnt this redundant or does it matter how the jax.sparse deals with synapses?

tests/test_api_equivalence.py
jnsbck, line 1: Could add test jaxley sparse vs. jax.sparse where all branches have same number of comps. Might be overkill though.

jnsbck: also, while implicitly tested via the integration currently `comp_edges_to_indices` and `convert_to_csc` are not covered explicitely. Dunno if it makes sense to write a seperate test for them tho

jaxley/solver_voltage.py
```diff
@@ -196,8 +197,50 @@ def step_voltage_implicit(
...
+    constant_terms,
+    axial_conductances,
+    data_inds,
+    indices,
+    indptr,
+    sinks,
+    delta_t,
+    n_nodes,
+    internal_node_inds,
+):
```
jnsbck, line 215: Type hints?

jnsbck: or did you leave them on purpose

jnsbck (changes requested):
Really cool! Looks great already! I left a few comments. Feel free to merge afterwards if tests pass. 

Overall, we should make sure that we are really clear about in what format we want to store all the attrs in `Module` and which of them we want to hide/expose for which reasons? I.e. should we hide comp_edges because of the "non-existant" branch points, but expose a version, i.e. via a property that does only return the existing ones. Should we do the same with branch_edges? Also we might want to bundle attributes which are only used together. I.e. if `jax_spsolve` is the only function that requires certain attrs, I'd be in favour of keeping keeping them together, as to not litter the `Module` class with a ton of arguments. I think it might become hard to track at some point. Also assuming that only one solver is used at a time, variable names could potentially be used for both jax_spsolve and  jaxley_spsolve (if sensible of course). 

Would be happy drafting a solid general purpose implementation of a sparse solver at some point as well. I think for example computing branch_edges from comp_edges would make branch_edges redundant and could be done by the solver if required.

Happy to give additional feedback if necessary. Lemme know. Otherwise good to go from my side after comments are adressed.

tests/test_api_equivalence.py
michaeldeistler: It's just to make very sure :)

jaxley/modules/branch.py
```diff
@@ -78,19 +85,13 @@ def __init__(
...
-        self.child_belongs_to_branchpoint = np.asarray([]).astype(int)
-        self.par_inds = np.asarray([]).astype(int)
-        self.total_nbranchpoints = 0
-        self.branchpoint_group_inds = np.asarray([]).astype(int)
-
-        self.children_in_level = []
-        self.parents_in_level = []
+        self.par_inds, self.child_inds, self.child_belongs_to_branchpoint = (
+            compute_children_and_parents(self.branch_edges)
+        )
```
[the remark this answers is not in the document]
michaeldeistler: These things are used at every step, so I would prefer to store them as `jnp.ndarray`s. We can of course make it a `Dict[jnp.ndarray]`.

jaxley/modules/compartment.py
```diff
@@ -76,6 +100,9 @@ def init_conds(self, params):
             "branch_diags": jnp.asarray([0.0]),
         }
 
+    def _init_conds_jax_spsolve(self, params):
```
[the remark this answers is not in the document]
michaeldeistler: Yes, this is happening in the follow-up PR now :)

jaxley/modules/network.py
```diff
@@ -127,28 +135,108 @@ def __getattr__(self, key: str):
...
+            [
+                jnp.asarray([0]),
+                jnp.cumsum(jnp.asarray([cell.cumsum_nseg[-1] for cell in self.cells])),
+            ]
+        )
+
+        self._comp_edges = pd.DataFrame()
+
+        # Add all the internal nodes.
+        for offset, cell in zip(self._cumsum_nseg_per_cell, self.cells):
```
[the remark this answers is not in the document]
michaeldeistler: It would make the ordering in `comp_edges` less intuitive. And I think if we sort afterwards we would have to be very careful to not mess up within-brach-sorting. I would prefer to keep it as it is for now.

jaxley/solver_voltage.py
michaeldeistler: Agreed!

jaxley/utils/cell_utils.py
michaeldeistler: TBH I find this a bit less readable

jaxley/utils/cell_utils.py
michaeldeistler: I find it a bit less readable.

jaxley/solver_voltage.py
michaeldeistler: Added them back in

tests/test_api_equivalence.py
michaeldeistler: Yes I added a lot of new tests in the new PR.

PR #432, opened 2024-09-27
Plotting of comp morphology

jnsbck:
I got side-tracked while working on a different PR and implemented plotting of the actual compartment structure:

The compartments are rendered in 3D and then the correct projection of the 3d mesh is plotted on whatever plane. 
```
fig, ax = plt.subplots(2, 3, figsize=(10, 5))

for i, orientation in enumerate([np.array([1,1,2]), np.array([0,1,0.1]), np.array([0,0.1,1])]):
    plot_cylinder_projection(orientation, length=5, radius=1, center=np.array([0,0,0]), dims=[0,1], ax=ax[0,i])
    plot_cylinder_projection(orientation, length=5, radius=1, center=np.array([0,0,0]), dims=[1,2], ax=ax[1,i])
    ax[0,i].set_title(f"Orientation: {orientation}")
plt.show()
```
image

Hence the compartment radii and lengths are correctly reflected in the plot.
image

However, plotting an imported morphology looked weird for some reason.
image

While looking into this, I noticed two things:
Since the cylindrical compartments are straight, they run from one compartment edge to the other on a straight line. However, they model a potentially very zigzaggy branch/path structure. This means:
a) Assuming the comp center is at loc=0.5 along the path (currently how we do it), can put the comp center off axis. 
b) Setting the comp length to the path length can result in comps which appear longer than the distance between their two endpoints.

This can be illustrated by adding a bit of noisy to the xyz coordinates of the example above.
image

Changing the compartment centers to be the center of the cylinder, already produces much better looking plots.
image

While I dont think changing b) makes sense (apart from for visualization maybe) I think changing a) might be a good idea.

Lemme know your thoughts.

_PR is missing documentation, some cleanup and handling of single point cases._

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1712,6 +1712,8 @@ class View:
         Args:
             ax: An axis into which to plot.
             col: The color for all branches.
+            type: Whether to plot as points ("scatter"), a line ("line") or the
+                actual volume of the compartment("volume").
             dims: Which dimensions to plot. 1=x, 2=y, 3=z coordinate. Must be a tuple of
                 two of them.
             morph_plot_kwargs: Keyword arguments passed to the plotting function.
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -33,8 +33,8 @@ from jaxley.utils.cell_utils import (
 )
 from jaxley.utils.debug_solver import compute_morphology_indices
 from jaxley.utils.misc_utils import childview, concat_and_ignore_empty
-from jaxley.utils.solver_utils import convert_to_csc
 from jaxley.utils.plot_utils import plot_comps, plot_morph
+from jaxley.utils.solver_utils import convert_to_csc
 
 
 class Module(ABC):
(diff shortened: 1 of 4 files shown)


michaeldeistler:
Hi Jonas, thanks, this can be super useful, especially for the radiuses.

I don't understand this sentence: `Assuming the comp center is at loc=0.5 along the path (currently how we do it), can put the comp center off axis.`

Afaik we currently have no notion of "comp center", so I don't understand what you mean.

jnsbck:
We do in '_update_nodes_with_xyz'.

michaeldeistler:
Okay! Then yes, I think I would be pro `changing the compartment centers to be the center of the cylinder`.

jnsbck:
Ready for review. Now `cell.vis(type="volume")` plots the compartments as seen below.
image
Not sure if `"volume"` is the best kwarg tho.

Q: Should I also add `col` kwarg for color like vis?

jaxley/modules/base.py
```diff
@@ -1226,6 +1240,10 @@ def _vis(
         morph_plot_kwargs: Dict,
     ) -> Axes:
         branches_inds = view["branch_index"].to_numpy()
+
+        if type == "volume":
+            return plot_comps(self[:], dims, ax, **morph_plot_kwargs)
```
michaeldeistler, line 1245: do we need the `[:]`?

jaxley/utils/plot_utils.py
```diff
@@ -38,3 +43,182 @@ def plot_morph(
...
+        center: The x,y,z coordinates of the center of the cylinder.
+        dims: The dimensions to project the cylinder onto, i.e. [0,1] xy-plane.
+        ax: The matplotlib axis to plot on.
+
+    Returns:
+        Plot of the cylinder projection.
+    """
+    if ax is None:
+        fig = plt.figure()
+        ax = fig.add_subplot(111)
```
michaeldeistler, line 126: `figsize=(3,3)` for consistency with `plot_morph()`

michaeldeistler (commented):
Thanks, awesome! Please add a test (just ensuring that there is no error) for a compartment, branch, cell, network, swc file, and `net.cell(1).vis(...)`

jaxley/modules/base.py
jnsbck: Input arg needs to be View. So [:] takes care of that if it's already a view it doesn't matter

jnsbck:
Thanks for the feedback!

Should I also add the col kwarg for consistency? https://github.com/jaxleyverse/jaxley/pull/432#issuecomment-2383240406
@michaeldeistler ?

michaeldeistler:
Yes I think it would be great to have a `col` argument!

jnsbck:
@michaeldeistler added tests and addressed all comments. Feel free to merge if you approve.

michaeldeistler (approved):

jnsbck:
swc files to reproduce the effect of zigzaging morphologies on the comp lengths. file extension changed to txt since swc not supported

[noisy_test.txt](https://github.com/user-attachments/files/17311398/noisy_test.txt)
[test.txt](https://github.com/user-attachments/files/17311399/test.txt)

PR #436, opened 2024-09-29
Allow changing nseg after initialization

michaeldeistler:
This PR is a follow-up to #418 and #426. It allows to set `nseg` **after** initialization of the module. This is achieved with a new `set_ncomp` method. This PR is the final requirement for a new release with these features.

### API

Works for branches:
```python
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=2)
branch.set_ncomp(4)
```

...for branches within a cell:
```python
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=2)
cell = jx.Cell(branch, parents=[-1, 0, 0])
cell.branch(1).set_ncomp(4)
```

...and for branches read from SWC files, in which case it reuses the xyzr to compute optional radiuses:
```python
cell = jx.read_swc(fname, nseg=2)
cell.branch(1).set_ncomp(4)
```

### Assumption and `Raises`
The `set_ncomp` function will raise an error if:  
- There are stimuli in **any** compartment in the module  
- There are recordings in **any** compartment in the module  
- The channels of the compartments are not the same within the branch that is modified  
- The lengths of the compartments are not the same within the branch that is modified 

In addition, unlike the data was read from an SWC file, `.set_ncomp()` raises an error if:  
- The radiuses of the compartments are not the same within the branch that is modified  

Finally, we currently raise `NotImplementedError` if the `Module` is a `Network`, e.g. for
```python
net.cell(0).branch(1).set_ncomp(4)
```
This is because I do not want to deal with synapse indexing for now.

We also raise a `NotImplementedError` if the `Module` is an entire `Cell`, e.g. for
```python
cell.set_ncomp(4)
```

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -5,6 +5,7 @@ import inspect
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from typing import Callable, Dict, List, Optional, Tuple, Union
+from warnings import warn
 
 import jax.numpy as jnp
 import numpy as np
@@ -32,9 +33,14 @@ from jaxley.utils.cell_utils import (
     v_interp,
 )
 from jaxley.utils.debug_solver import compute_morphology_indices
-from jaxley.utils.misc_utils import childview, concat_and_ignore_empty
+from jaxley.utils.misc_utils import (
+    childview,
+    concat_and_ignore_empty,
+    cumsum_leading_zero,
+)
 from jaxley.utils.plot_utils import plot_comps, plot_morph
 from jaxley.utils.solver_utils import convert_to_csc
+from jaxley.utils.swc import build_radiuses_from_xyzr
 
 
 class Module(ABC):
@@ -109,6 +115,7 @@ class Module(ABC):
 
         # x, y, z coordinates and radius.
         self.xyzr: List[np.ndarray] = []
+        self._radius_generating_fns = None  # Defined by `.read_swc()`.
 
         # For debugging the solver. Will be empty by default and only filled if
         # `self._init_morph_for_debugging` is run.
@@ -156,6 +163,14 @@ class Module(ABC):
     def __dir__(self):
         base_dir = object.__dir__(self)
         return sorted(base_dir + self.synapse_names + list(self.group_nodes.keys()))
+
+    @property
+    def _module_type(self):
+        """Return type of the module (compartment, branch, cell, network) as string.
+
+        This is used to perform asserts for some modules (e.g. network cannot use
+        `set_ncomp`) without having to import the module in `base.py`."""
+        return self.__class__.__name__.lower()
 
     def _append_params_and_states(self, param_dict: Dict, state_dict: Dict):
         """Insert the default params of the module (e.g. radius, length).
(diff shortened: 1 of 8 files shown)


jaxley/modules/cell.py
```diff
@@ -203,17 +210,26 @@ def _init_morph_jax_spsolve(self):
...
         self._comp_edges = pd.concat(
             [
-                [offset, offset, 0] + branch._comp_edges
-                for offset, branch in zip(self.cumsum_nseg, self.branch_list)
+                pd.DataFrame().from_dict(
+                    {
+                        "source": list(range(cumsum_nseg, nseg - 1 + cumsum_nseg))
+                        + list(range(1 + cumsum_nseg, nseg + cumsum_nseg)),
+                        "sink": list(range(1 + cumsum_nseg, nseg + cumsum_nseg))
+                        + list(range(cumsum_nseg, nseg - 1 + cumsum_nseg)),
```
jnsbck, line 220: do we want lists specifically or isnt it better to replace `list(range(...))` by `np.arange`

jaxley/modules/cell.py
```diff
@@ -203,17 +210,26 @@ def _init_morph_jax_spsolve(self):
...
-        # `branch_list` is not needed anymore because all information it contained is
-        # now also in `self._comp_edges`.
-        del self.branch_list
+        self._comp_edges["type"] = 0
+
+        # Casting source and sink to int. This is necessary for branches which have
+        # just a single compartment, because in these cases, `_comp_edges` is empty
+        # and is instantiated as `float` by default.
+        self._comp_edges["source"] = self._comp_edges["source"].astype(int)
+        self._comp_edges["sink"] = self._comp_edges["sink"].astype(int)
```
jnsbck, line 232: unless you instantiate the `_comp_edges` dataframe with `dtype=int` from the get go no?

tests/test_set_ncomp.py
```diff
@@ -0,0 +1,194 @@
...
+import pytest
+
+import jaxley as jx
+from jaxley.channels import HH
+
+
+@pytest.mark.parametrize(
+    "property", ["radius", "capacitance", "length", "axial_resistivity"]
+)
+def test_raise_for_uneven_modules(property):
```
jnsbck, line 22: uneven is a bit missleading imo. Id prefer sth like heterogenic, irregular

jaxley/modules/cell.py
```diff
@@ -85,28 +86,35 @@ def __init__(
...
+        self.total_nbranches = len(branch_list)
+        self.nbranches_per_cell = [len(branch_list)]
+        self.comb_parents = jnp.asarray(parents)
+        self.comb_children = compute_children_indices(self.comb_parents)
+        self.cumsum_nbranches = jnp.asarray([0, len(branch_list)])
+
+        # Compartment structure. These arguments have to be rebuilt when `.set_ncomp()`
+        # is run.
+        self.nseg_per_branch = jnp.asarray([branch.nseg for branch in branch_list])
         self.nseg = int(jnp.max(self.nseg_per_branch))
```
jnsbck, line 98: should we rename this to `max_nseg`? or make this an array?

jaxley/modules/branch.py
```diff
@@ -146,6 +148,51 @@ def _init_morph_jax_spsolve(self):
...
+            list(chain(*[c.channel_states for c in self.channels])),
+            self._radius_generating_fns,
+            min_radius,
+        )
+
+        # Update the morphology indexing (e.g., `.comp_edges`).
+        self.initialize()
+
 
 class BranchView(View):
```
jnsbck, line 197: does branchview have a method / arg to check the number of segments? 
I think it would be nice if `cell.branch(0).nseg` or similar would show the correct number.

jaxley/modules/branch.py
```diff
@@ -167,3 +214,60 @@ def __getattr__(self, key):
...
+            - When there are stimuli in any compartment in the module.
+            - When there are recordings in any compartment in the module.
+            - When the channels of the compartments are not the same within the branch
+            that is modified.
+            - When the lengths of the compartments are not the same within the branch
+            that is modified.
+            - Unless the morphology was read from an SWC file, when the radiuses of the
+            compartments are not the same within the branch that is modified.
+        """
+        if self.pointer.module_type == "network":
```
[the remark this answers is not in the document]
michaeldeistler: We would have to enforce empty synapses. But yes, we could do this. I will create a follow-up issue that describes all limitations of `set_ncomp()`.

jaxley/modules/cell.py
michaeldeistler: Yes I agree but let's do this in another PR.

jaxley/modules/branch.py
michaeldeistler: I am not sure if we need this---`cell.branch(0).show()` will have `nseg` rows, I think it would be fine for the user to just do this (as compared to having another magical argument which the user has to learn how to use).

jaxley/modules/branch.py
```diff
@@ -56,7 +58,7 @@ def __init__(
             compartment_list = compartments
 
         self.nseg = len(compartment_list)
-        self.nseg_per_branch = [self.nseg]
+        self.nseg_per_branch = jnp.asarray([self.nseg])
```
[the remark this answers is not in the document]
michaeldeistler: I want to move towards making everything `jnp` to avoid potential overhead with copying between CPU and GPU (we are very inconsistent with this atm)

jaxley/modules/base.py
```diff
@@ -381,6 +384,137 @@ def _data_set(
...
+        # 1) Want to maintain the total length of a branch.
+        # 2) Want to use the SWC inferred radius.
+        #
+        # Compute new compartment lengths.
+        comp_lengths = np.sum(compartment_lengths) / ncomp
+        view["length"] = comp_lengths
+
+        # Compute new compartment radiuses.
+        if radius_generating_fns is not None:
+            view["radius"] = build_radiuses_from_xyzr(
```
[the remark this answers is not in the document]
michaeldeistler: yes we use it in the swc parser too :)

jaxley/modules/base.py
```diff
@@ -381,6 +384,137 @@ def _data_set(
...
+        if not (view[channel_names].var() == 0.0).all():
+            raise ValueError(
+                "Some channel exists only in some compartments of the branch which you"
+                "are trying to modify. This is not allowed. First specify the number"
+                "of compartments with `.set_ncomp()` and then insert the channels"
+                "accordingly."
+            )
+
+        if not (view[channel_param_names + channel_state_names].var() == 0.0).all():
+            raise ValueError(
```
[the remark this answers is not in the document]
michaeldeistler: hmmm not sure. I think I like having the option to already have channels in the module. In the future, I want to move towards allowing __more__ things before `set_ncomp`, in particular stimuli and recordings.

jaxley/modules/cell.py
michaeldeistler: using + to concat the lists is handy here :)

PR #437, opened 2024-10-03
more options for plotting

jnsbck:
I have continued #432 and added two new options:

- plotting the true traced morphology (i.e. including radii of every traced point not just for each comp) 
image

- and allowing to plot all 3 axes in a 3d plot
image

Also added this to the tests. They are passing.


michaeldeistler (approved):
Love it! A few code-health comments below. Good to go once addressed!

We should also write a small tutorial about all of these awesome features (maybe as part of the SWC tutorial?)

Issue #377, opened 2024-07-12
Allow `cell.comp(0)` or `net.branch(0)` or `net.comp(0)`


jnsbck:
Should we allow `cell.comp(0)` or `net.branch(0)` or `net.comp(0)`? I think this would be helpful sometimes to be able to reference comps by their global index that can also be found in `cell.nodes`. Currently this can be achieved via cell.branch.comp(0), which I think might be an artefact and is less clear.

Issue #354, opened 2024-04-24
refactor `modules/base.py` to make `View` behave like `Module`


jnsbck:
>In the long term, it would be nice to make `View` behave more like `Module` without all of these properties essentially just indexing into `View.pointer` in some way. Imo most methods should be part of module if possible and only passed through to view, with view only showing what's in scope. While a method in module acts on the entire thing, in view it only acts on whatever is in view. This could be done by having a `self.viewed_indexes` / `self.in_scope` attr in Module, that decides what is `set`, where things are `insert`ed etc. This would get rid of a lot of complexity and duplicate methods. For example would this make many hidden methods, like `_record` in Module obsolete. Additionally `view` can just be a property that returns `View.nodes`.

As suggested in #351

jnsbck:
This was adressed with #447

Issue #338, opened 2024-04-12
Behavior of subclassing groups differs between `net` and `cell`.


michaeldeistler:
If one subclasses groups of `net` then one needs to use the **global** index of the cell (i.e. in the entire network):
```python
net = jx.Network([cell for _ in range(5)])
net.cell(3).add_to_group("exc")
net.exc.cell(3).set(...)
```

Contrary to that, for `cell`, one has to use the local index of the branch within the group:
```python
cell = jx.Cell(Branch, [-1, 0, 0])
cell.branch(2).add_to_group("exc")
cell.exc.branch(0).set(...)
```

michaeldeistler:
I think this can be closed, right @jnsbck ?

jnsbck:
Should have been addressed in #447, will add a test for this in #453

Issue #291, opened 2024-03-15
better use of local indices


jnsbck:
(1) the function `index_of_loc` returns the highest index for the closest segment and the lowest for the furthest, see #30 and (2) it is multiplying with branch_ind because we had to rely on global compartment indices -- now this could be solved with local indices and then it is not necessary anymore to pass a branch_ind to the function
https://github.com/jaxleyverse/jaxley/blob/729c1cafdb80ab66f8c5013fbee26d7abc042613/jaxley/utils/cell_utils.py#L161

_Originally posted by @lappalainenj in https://github.com/jaxleyverse/jaxley/issues/277#issuecomment-1999963966_
            
Also see [here](https://github.com/jaxleyverse/jaxley/pull/277#issuecomment-1999672863)

michaeldeistler:
Can this be closed @jnsbck ?

jnsbck:
Should be adressed by #447.

PR #453, opened 2024-10-23
Continuing better Viewing (#447)

jnsbck:
The following is a copy of the remaining todos from #447

> Merging this will make https://github.com/jaxleyverse/jaxley/pull/329, https://github.com/jaxleyverse/jaxley/pull/330, https://github.com/jaxleyverse/jaxley/pull/351 obsolete and close https://github.com/jaxleyverse/jaxley/issues/377, https://github.com/jaxleyverse/jaxley/issues/354, probably https://github.com/jaxleyverse/jaxley/issues/338 and I guess also https://github.com/jaxleyverse/jaxley/issues/133
> Will also make https://github.com/jaxleyverse/jaxley/issues/264 and https://github.com/jaxleyverse/jaxley/issues/134 trivial
> Also might wanna revisit https://github.com/jaxleyverse/jaxley/issues/291 after this

> ### Comments:
> - Moved `_append_multiple_synapses` and associated functionality to `Network`.
> - `SynapseView` replaced by `module.TestSynapse.edge([0,1])`
> 
> ### Open Questions:
> - How should we name `self.view`, `self.base`, `self.at`?
> - How should we handle local indices? Should they be recomputed dynamically? I.e.`self.comp([2,3,4]).comp(1)` or to allow groups to have a local index `self.group_name.comp(0)`?
> - change name of `_cells_in_view` to `_cell_indices` etc.
> - Treat synapse trainables, i.e. edges the same as nodes? which would also allow param sharing
> - Discuss what we want to allow with new View that we didnt before -> diff of NotImplementedError and other things that are now trivial
> - How to treat local inds? Should we just have single local inds col? This would mean doing net.cell([1,2]).comp(0) would only show the first comp of cell 1 rather than the first comp in every branch of cell 1 and 2
> - should controlled_by_param be 0 or np.arange(len(self.nodes)) by default?

### Immediate Todos:
- [x] Fix broken tests
- [x] Add new tests
- [x] Add documentation and type hints
~~- [x] update existing tutorials with new API (edge)~~ Will be removed -> #463 

### Smaller Todos
- [x] locals in edges to front and globals to the back
- [x] remove local indices from edges?
- [x] interplay of nodes and edges -> delegated to #463
- [x] groups are shared per default! -> controlled by param
- [x] add test for #338

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -476,28 +476,6 @@ class Module(ABC):
             View of the module at the specified edge index."""
         return self._at_edges("edge", idx)
 
-    def pre(self, idx: Any) -> View:
-        """Return a View of the module at the selected pre-synaptic compartments(s).
-
-        Args:
-            idx: index of the edge to view.
-
-        Returns:
-            View of the module filtered by the selected pre-comp index."""
-        # TODO: pre and post could just modify scope
-        #  -> self.scope=self.scope+"_pre" and then call edge?
-        return None  # self._at_edges("edge", idx)
-
-    def post(self, idx: Any) -> View:
-        """Return a View of the module at the selected post-synaptic compartments(s).
-
-        Args:
-            idx: index of the edge to view.
-
-        Returns:
-            View of the module filtered by the selected post-comp index."""
-        return None  # self._at_edges("edge", idx)
-
     def loc(self, at: Any) -> View:
         """Return a View of the module at the selected branch location(s).
 
@@ -623,7 +601,7 @@ class Module(ABC):
     @property
     def view(self):
         """Return view of the module."""
-        return View(self, self._nodes_in_view)
+        return View(self, self._nodes_in_view, self._edges_in_view)
 
     @property
     def _module_type(self):
@@ -1108,12 +1086,19 @@ class Module(ABC):
         return np.sqrt(np.sum((start_xyz - end_xyz) ** 2))
 
     def delete_trainables(self):
-        # TODO: MAKE THIS WORK FOR VIEW?
+        # TODO: Test that this correctly works for View!
         """Removes all trainable parameters from the module."""
-        assert isinstance(self, Module), "Only supports modules."
-        self.base.indices_set_by_trainables = []
-        self.base.trainable_params = []
-        self.base.num_trainable_params = 0
+
+        if isinstance(self, View):
+            trainables_and_inds = self._filter_trainables(is_viewed=False)
+            self.base.indices_set_by_trainables = trainables_and_inds[0]
+            self.base.trainable_params = trainables_and_inds[1]
+            self.base.num_trainable_params -= self.num_trainable_params
+        else:
+            self.base.indices_set_by_trainables = []
+            self.base.trainable_params = []
+            self.base.num_trainable_params = 0
+        self._update_view()
 
     def add_to_group(self, group_name: str):
         """Add a view of the module to a group.
@@ -1135,7 +1120,6 @@ class Module(ABC):
             )
 
     def get_parameters(self) -> List[Dict[str, jnp.ndarray]]:
-        # TODO: MAKE THIS WORK FOR VIEW?
         """Get all trainable parameters.
 
         The returned parameters should be passed to `jx.integrate(..., params=params).
@@ -1144,7 +1128,7 @@ class Module(ABC):
             A list of all trainable parameters in the form of
                 [{"gNa": jnp.array([0.1, 0.2, 0.3])}, ...].
         """
-        return self.base.trainable_params
+        return self.trainable_params
 
     def get_all_parameters(
         self, pstate: List[Dict], voltage_solver: str
@@ -1400,9 +1384,11 @@ class Module(ABC):
         self.base.debug_states["par_inds"] = self.base.par_inds
 
     def record(self, state: str = "v", verbose=True):
-        in_view = (
-            self._edges_in_view if state in self.edges.columns else self._nodes_in_view
-        )
+        in_view = None
+        in_view = self._edges_in_view if state in self.edges.columns else in_view
+        in_view = self._nodes_in_view if state in self.nodes.columns else in_view
+        assert in_view is not None, "State not found in nodes or edges."
+
         new_recs = pd.DataFrame(in_view, columns=["rec_index"])
         new_recs["state"] = state
         self.base.recordings = pd.concat([self.base.recordings, new_recs])
@@ -1413,11 +1399,27 @@ class Module(ABC):
                 f"Added {len(in_view)-sum(has_duplicates)} recordings. See `.recordings` for details."
             )
 
+    def _update_view(self):
+        """Update the attrs of the view after changes in the base module."""
+        if isinstance(self, View):
+            scope = self._scope
+            current_view = self._current_view
+            self.__dict__ = View(
+                self.base, self._nodes_in_view, self._edges_in_view
+            ).__dict__
+            self._scope = scope
+            self._current_view = current_view
+
     def delete_recordings(self):
-        # TODO: MAKE THIS WORK FOR VIEW?
         """Removes all recordings from the module."""
-        assert isinstance(self, Module), "Only supports modules."
-        self.base.recordings = pd.DataFrame().from_dict({})
+        if isinstance(self, View):
+            base_recs = self.base.recordings
+            self.base.recordings = base_recs[
+                ~base_recs.isin(self.recordings).all(axis=1)
+            ]
+            self._update_view()
+        else:
+            self.base.recordings = pd.DataFrame().from_dict({})
 
     def stimulate(self, current: Optional[jnp.ndarray] = None, verbose: bool = True):
         """Insert a stimulus into the compartment.
@@ -1566,18 +1568,26 @@ class Module(ABC):
         return (state_name, external_input, inds)
 
     def delete_stimuli(self):
-        # TODO: MAKE THIS WORK FOR VIEW?
         """Removes all stimuli from the module."""
-        assert isinstance(self, Module), "Only supports modules."
-        self.base.externals.pop("i", None)
-        self.base.external_inds.pop("i", None)
+        self.delete_clamps("i")
 
     def delete_clamps(self, state_name: str):
-        # TODO: MAKE THIS WORK FOR VIEW?
         """Removes all clamps of the given state from the module."""
-        assert isinstance(self, Module), "Only supports modules."
-        self.base.externals.pop(state_name, None)
-        self.base.external_inds.pop(state_name, None)
+        if state_name in self.externals:
+            keep_inds = ~np.isin(
+                self.base.external_inds[state_name], self._nodes_in_view
+            )
+            base_exts = self.base.externals
+            base_exts_inds = self.base.external_inds
+            if np.all(~keep_inds):
+                base_exts.pop(state_name, None)
+                base_exts_inds.pop(state_name, None)
+            else:
+                base_exts[state_name] = base_exts[state_name][keep_inds]
+                base_exts_inds[state_name] = base_exts_inds[state_name][keep_inds]
+            self._update_view()
+        else:
+            pass  # does not have to be deleted if not in externals
 
     def insert(self, channel: Channel):
         """Insert a channel into the module.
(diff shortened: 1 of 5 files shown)


jnsbck:
Added a bunch of tests, fixed a few bugs that popped up and generally cleaned up the code. Also integrated a few changes we discussed, like removing inds from the edges dataframe. `edge` is kept for now. Removing this should be part of #463

jaxley/modules/base.py
```diff
@@ -1515,19 +1517,27 @@ def _data_external_input(
...
-        self.base.external_inds.pop("i", None)
+        self.delete_clamps("i")
 
-    # TODO: MAKE THIS WORK FOR VIEW?
     def delete_clamps(self, state_name: str):
         """Removes all clamps of the given state from the module."""
-        assert isinstance(self, Module), "Only supports modules."
-        self.base.externals.pop(state_name, None)
-        self.base.external_inds.pop(state_name, None)
+        if state_name in self.externals:
```
michaeldeistler, line 1526: do we have tests for this?

jaxley/modules/base.py
```diff
@@ -2272,67 +2284,68 @@ def _set_externals_in_view(self):
...
-            if len(trainable_inds) > 0
-            else []
-        )
-        trainable_node_inds_in_view = np.intersect1d(
-            trainable_inds, self._nodes_in_view
-        )
+    def _filter_trainables(
+        self, is_viewed: bool = True
+    ) -> Tuple[List[np.ndarray], List[Dict]]:
+        """filters the trainables inside and outside of the view
```
michaeldeistler, line 2290: Sentence case please.

michaeldeistler: Also, could you add a sentence to the docstring about where this is used? I always find it hard to have these hidden methods where it is not immediately clear why they are needed.

jaxley/modules/base.py
```diff
@@ -2272,67 +2284,68 @@ def _set_externals_in_view(self):
...
-            }
-            trainable_params_in_view.append(syn_trainable_params_in_view)
-            syn_inds_set_by_trainables_in_view = self.base.indices_set_by_trainables[
-                idx
-            ][syn_inds_in_view]
-            índices_set_by_trainables_in_view.append(syn_inds_set_by_trainables_in_view)
+            índices_set_by_trainables_in_view.append(inds[completely_in_view])
+            partial_inds = inds[partially_in_view][in_view[partially_in_view]]
+
+            # the indexing above can lead to inconsistent shapes.
```
michaeldeistler, line 2325: Why? More detailed comment please.

jaxley/modules/base.py
```diff
@@ -2272,67 +2284,68 @@ def _set_externals_in_view(self):
...
+        indices_set_by_trainables = [
             inds for inds in índices_set_by_trainables_in_view if len(inds) > 0
         ]
-        self.trainable_params = [
+        trainable_params = [
             p for p in trainable_params_in_view if len(next(iter(p.values()))) > 0
         ]
+        return indices_set_by_trainables, trainable_params
+
+    def _set_trainables_in_view(self):
```
michaeldeistler, line 2342: docstring

tests/test_viewing.py
```diff
@@ -268,12 +256,11 @@ def test_solve_indexer():
 
 # make sure all attrs in module also have a corresponding attr in view
 @pytest.mark.parametrize("module", [comp, branch, cell, net])
-def test_view_attrs(module):
+def test_view_attrs(module: jx.Compartment | jx.Branch | jx.Cell | jx.Network):
```
michaeldeistler, line 259: docstring

michaeldeistler (commented):
Awesome! I did not review in a lot of detail, but my overall feedback would be to (1) write slightly more comprehensive docstrings (outlining also _why_ particular functions are needed) and (2) adding more detail to comments, as many are not immediately clear to me. Let's also add tests for removing recordings, stimuli, and trainables from views. Thanks!

jaxley/modules/base.py
```diff
@@ -179,7 +183,7 @@ def _childviews(self) -> List[str]:
     def __getitem__(self, index):
         supported_lvls = ["network", "cell", "branch"]  # cannot index into comp
 
-        # TODO: SHOULD WE ALLOW GROUPVIEW TO BE INDEXED?
+        # TODO FROM #447: SHOULD WE ALLOW GROUPVIEW TO BE INDEXED?
```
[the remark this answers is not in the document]
jnsbck: The problem is that for Modules, i.e. ´net[0,0,0]` it is straightforward to lazy index into by just calling `net.cell(0).branch(0).comp(0)` internally, likewise for `net.cell(0)[0,0]`. However, for `net.cell0` this is not the case, since we cannot tell easily what the user is indexing into. I.e. they might do `net.cell0` or `net.comp0`, which are totally different, but equivalent on the group level. Now should we allow them to still be able to do `net.comp0[0,0,0]`? 

Decided to disable it for now and only allow indexing into Modules and Views thereof. I think that is sensible. The user can of course still index into groups however they like, they just have to do it via `comp`, `cell`, `branch` instead of via `__getitem__`.

jaxley/modules/base.py
```diff
@@ -41,6 +42,23 @@
 from jaxley.utils.swc import build_radiuses_from_xyzr
 
 
+def only_allow_module(func):
```
jnsbck, line 45: I added a decorator that we can add to functions that cannot be run on Views (yet).
I hope I did not miss any.

jaxley/modules/base.py
```diff
@@ -49,7 +67,42 @@ class Module(ABC):
 
     Modules can be traversed and modified using the `at`, `cell`, `branch`, `comp`,
     `edge`, and `loc` methods. The `scope` method can be used to toggle between
-    global and local indices.
+    global and local indices. Traversal of Modules will return a `View` of itself,
```
jnsbck, line 70: Added more doc to `Module`

jaxley/modules/base.py
```diff
@@ -227,39 +286,37 @@ def reindex_a_by_b(
...
+        idcs = reindex_a_by_b(idcs, global_idx_cols[0])
+        idcs = reindex_a_by_b(idcs, global_idx_cols[1], global_idx_cols[0])
+        idcs = reindex_a_by_b(idcs, global_idx_cols[2], global_idx_cols[:2])
+        idcs.columns = [col.replace("global", "local") for col in global_idx_cols]
+        self.nodes[local_idx_cols] = idcs[local_idx_cols].astype(int)
 
         # move indices to the front of the dataframe; move controlled_by_param to the end
+        # move indices of current scope to the front and the others to the back
+        not_scope = "global" if self._scope == "local" else "local"
+        self.nodes = reorder_cols(
```
jnsbck, line 303: If scope is local, local_inds will be shown at the front, globals at the back and vice versa.

jaxley/modules/base.py
```diff
@@ -413,6 +470,8 @@ def _at_nodes(self, key: str, idx: Any) -> View:
 
         Keys can be `cell`, `branch`, `comp` and determine which index is used to filter.
         """
+        base_name = self.base.__class__.__name__
+        assert self.base._has_childview(key), f"{base_name} does not support {key}."
```
jnsbck, line 474: Disallow stuff like `comp.cell(0)`. While I think there is technically no harm in allowing it  (the cell_index also shows up in `nodes`), it makes little sense and could lead to more problems than it solves for the user, i.e. them calling `comp.cell(1)`, which does not exist.

jaxley/modules/base.py
```diff
@@ -613,17 +650,18 @@ def copy(
         Returns:
             A part of the module or a copied view of it."""
         view = deepcopy(self)
-        # TODO: add reset_index, i.e. for parents, nodes, edges etc. such that they
+        warnings.warn("This method is experimental, use at your own risk.")
```
jnsbck, line 653: Add warning, since this only takes a copy of View.

jaxley/modules/base.py
```diff
@@ -1200,14 +1245,13 @@ def get_all_parameters(
...
             # Instead edges should work similar to nodes (would also allow for
             # param sharing).
+            synapse_inds = self.base.edges.groupby("type").rank()["global_edge_index"]
+            synapse_inds = (synapse_inds.astype(int) - 1).to_numpy()
             if key in self.base.synapse_param_names:
-                syn_name_from_param = key.split("_")[0]
-                syn_edges = self.__getattr__(syn_name_from_param).edges
-                inds = syn_edges.loc[inds.reshape(-1)]["local_edge_index"].values
-                inds = inds.reshape(-1, 1)
+                inds = synapse_inds[inds]
```
jnsbck, line 1254: fixes #467, although still need to add tests for this.

jaxley/modules/base.py
```diff
@@ -2090,14 +2164,19 @@ def move_to(
                 "NaN coordinate values detected. Shift amounts cannot be computed. Please run compute_xyzr() or assign initial coordinate values."
             )
 
-        root_xyz_cells = np.array([c.xyzr[0][0, :3] for c in self.cells])
+        # can only iterate over cells for networks
+        # lambda makes sure that generator can be created multiple times
+        base_is_net = self.base._current_view == "network"
```
jnsbck, line 2169: added special case, since disallowing cell.cells breaks this

tests/test_viewing.py
```diff
@@ -304,7 +300,331 @@ def test_view_attrs(module):
...
-# add cell.comp (branch is skipped also for param sharing)
-# add tests for new features i.e. iter, context, scope
+comp = jx.Compartment()
+branch = jx.Branch([comp] * 4)
+cell = jx.Cell([branch] * 4, parents=[-1, 0, 0, 0])
+net = jx.Network([cell] * 4)
+
+
+@pytest.mark.parametrize("module", [comp, branch, cell, net])
+def test_view_supported_index_types(module):
```
jnsbck, line 310: test all types supported to index into Modules

tests/test_viewing.py
```diff
@@ -304,7 +300,331 @@ def test_view_attrs(module):
...
+
+            # for loc test float and list of floats
+            assert module.loc(0.0), "Failed for float"
+            assert module.loc([0.0, 0.5, 1.0]), "Failed for List[float]"
+    else:
+        with pytest.raises(AssertionError):
+            module.comp(0)
+
+
+def test_select():
```
jnsbck, line 340: Add test for select functionality

tests/test_viewing.py
```diff
@@ -304,7 +300,331 @@ def test_view_attrs(module):
...
+
+    # convert to dict so order of cols and index dont matter for __eq__
+    assert nodes1.to_dict() == nodes2.to_dict()
+    assert np.all(nodes_control_param1 == 0)
+    assert np.all(nodes_control_param2 == nodes2["global_cell_index"])
+
+    assert np.all(edges1 == edges2)
+
+
+def test_view_equals_module():
```
jnsbck, line 584: add test to check if `comp` and `branch.comp(0)` behave the same for recordings, setting params and stimulating

jaxley/modules/base.py
michaeldeistler: Great!

jaxley/modules/base.py
michaeldeistler: great! Yes, let's disallow it.

jaxley/modules/base.py
```diff
@@ -41,6 +42,23 @@
 from jaxley.utils.swc import build_radiuses_from_xyzr
 
 
+def only_allow_module(func):
+    """Decorator to only allow the function to be called on Module instances.
```
michaeldeistler, line 46: hm maybe this should live in the `utils` folder?

michaeldeistler (approved):
Amazing! Good to go (one minor comment below)!

jaxley/modules/base.py
jnsbck: I also thought this, but wanted to avoid circular import issues, since `isinstance(self, View)` is used, which means we would have to import `View` in utils. Could work around this by using `self.__class__.__name__ == "View"` but I dislike this a bit. Lemme know your thoughts @michaeldeistler

michaeldeistler: ah, ok...let us keep it where it is then (but maybe move to the bottom of the file?)

jnsbck: Also does not work, since it's used in Module and needs to be defined above :/

jnsbck: One option would be to fold it into `Module`. Will merge it like it is, feel free to change this if you dislike it.

Issue #343, opened 2024-04-16
`softplus` transform does not properly rescale parameters


michaeldeistler:
### Problem

The `ParamTransform` transforms params to be unbounded and (as a nice side effect) roughly z-scores them. However, if one an upper **or** lower bound are passed, then it does not z-score.

### Solution

We should generally find a good solution to z-score bounded, unbounded, and half-bounded parameters.

Maybe sth like the following:
```python
tf: Dict = {}

# Some parameters are bounded from both sides.
tf = tf | {"HH_gNa": jx.SigmoidTransform(lower=0.0, upper=2.0)}
tf = tf | {"HH_gK": jx.SigmoidTransform(lower=0.0, upper=1.0)}

# Some parameters are bounded from above.
tf = tf | {"length": jx.SoftplusTransform(upper=100.0, scale=30.0)}

# Some parameters are bounded from below.
tf = tf | {"radius": jx.SoftplusTransform(lower=0.0, scale=6.0)}

# Some parameters are unbounded but should be z-scored.
tf = tf | {"IonotropicSynapse_gS": jx.AffineTransform(scale=5.0, offset=2.0)}

transform = jx.ParamTransform(tf)

# Apply transform
params = net.get_parameters()
opt_params = transform.inverse(params)
```

michaeldeistler:
@manuelgloeckler you might be interested in this?

PR #456, opened 2024-10-23
Updated transforms base

manuelgloeckler:
# Transform API Changes

This pull request introduces modifications to the `ParamsTransform` API. Now, transforms must be explicitly specified for each parameter, ensuring more precise control over their application. Additionally, the structure of the transforms (such as PyTrees or dictionaries) must match the structure of the parameters themselves. For further details, please refer to issue #343.

## Why These Changes?

The primary goal of these changes is to give *and* enforce manual control of reparameterizations to the User. Previously this was happening automatically, but suboptimal. 

## Example Usage

Below is an example demonstrating how to apply the updated API with custom transforms to various cell parameters in a `jaxley` model:

```python
import jaxley as jx
import jaxley.optimize.transforms as jt

# Define a model with compartments and branches
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=2)
cell = jx.Cell(branch, parents=[-1, 0, 0])

# Specify which parameters to make trainable
cell.branch("all").make_trainable("radius")
cell.branch(2).make_trainable("radius")
cell.branch(1).make_trainable("length")
cell.branch(0).make_trainable("radius")
cell.branch("all").comp("all").make_trainable("axial_resistivity")
cell.make_trainable("capacitance")

# Retrieve the parameters

# Define a function to create appropriate transforms for each parameter
def create_transform(name):
    if name == "axial_resistivity":
        # Must be positive; apply Softplus and scale to match initialization
        return jt.ChainTransform([jt.SoftplusTransform(0), jt.AffineTransform(5000, 0)])
    elif name == "length":
        # Apply Softplus and affine transform for the 'length' parameter
        return jt.ChainTransform([jt.SoftplusTransform(0), jt.AffineTransform(10, 0)])
    else:
        # Default to a Softplus transform for other parameters
        return jt.SoftplusTransform(0)

# Apply the transforms to the parameters
transforms = [{k: create_transform(k) for k in param} for param in params]
tf = jt.ParamTransform(transforms)

# Obtain the unconstrained parameters
params_unconstrained = tf.inverse(params)
```

## Key Benefits

- **Explicit Transformations**: This change makes it clear which transformation is applied to each parameter, increasing transparency and control.


jaxley/optimize/transforms.py
```diff
@@ -1,31 +1,153 @@
...
+from jax import Array
+from jax.typing import ArrayLike
 
 from jaxley.solver_gate import save_exp
 
 
-def sigmoid(x: jnp.ndarray) -> jnp.ndarray:
-    """Sigmoid."""
-    return 1 / (1 + save_exp(-x))
+class Transform(ABC):
```
michaeldeistler, line 14: can it have a `def forward`? And the forward `__call__()` just runs `forward()`?

tests/test_transforms.py
```diff
@@ -8,71 +8,127 @@
...
+    params = [{"param_array_1": jnp.asarray(np.linspace(-1, 1, 4))}]
+
+    tf = ParamTransform(tf_dict)
+
+    forward = tf.forward(params)
+    inverse = tf.inverse(forward)
+
+    assert np.allclose(
+        inverse[0]["param_array_1"], params[0]["param_array_1"]
+    ), f"{transform} forward, inverse failed."
```
michaeldeistler, line 134: Awesome! Can we have some more elaborate test that checks whether it plays well with `.get_parameters()`? 

E.g.
```python
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=2)
cell = jx.Cell(branch, parents=[-1, 0, 0])

cell.branch("all").make_trainable("radius")
cell.branch(2).make_trainable("radius")
cell.branch(1).make_trainable("length")
cell.branch(0).make_trainable("radius")
cell.branch("all").comp("all").make_trainable("axial_resistivity")
cell.make_trainable("capacitance")

transform = ...

transformed_params = transform.forward(params)
recovered_params = transform.inverse(transformed_params)
# ...test that they are still the same, or that transformed_params has no NaN.
```

michaeldeistler (commented):
Looks awesome!

Two things:
1) Could you write a more explicit PR description that shows the API change?
2) I suggest to add a slightly more elaborate test, see below.

Thanks!

jaxley/optimize/transforms.py
manuelgloeckler: Sure. Actually then also ParamsTransform can subclass it?

tests/test_transforms.py
manuelgloeckler: Added it.

I do scale everything by 1/1000 (as axial resistivity is 5000 for which most transforms are not numerically invertible)

manuelgloeckler:
Alright, I did add an API description with an example.

There might be a few other considerations:
- We could name the transforms by what they are doing and not what (i.e. Sigmoid -> Bound, Softplus -> BoundLower). It would make it easier to use, but also less visible what actually is happening.
- We could provide a "default reparameterization" i.e. basically a dict with all the names that can come up in jaxley and provide a sensible reparameterization (i.e. basically what the `create_transform` is doing in the example)

Matthijspals:
I updated the tutorial notebook with the new parameter transform function. I think (at a later point) it would be nice to have a longer section (or separate tutorial) to describe how to use them!

Note that the jit function in the notebook is commented out at the moment, see Issue #467 
(so it probably makes sense to wait with merging until that issue is fixed, and the notebook is updated again)

michaeldeistler (approved):

PR #458, opened 2024-10-23
Reduce warnings

manuelgloeckler:
# Fixing deprecation warnings
We change all:
- jax.tree_map -> jax.tree_util.tree_map
- Pytest ignores TestSynapse

Note: We do not change use of `a_max` in `jnp.clip`, because this require jax>0.4.30 (which is too recent yet)


manuelgloeckler:
There are quite a few unused imports in each file which I could additionally autofix with ruff in this PR.

michaeldeistler (approved):

PR #459, opened 2024-10-23
Make several methods private; delete unused methods

### LINKED ISSUE #448: Make some methods of `Module` private
Currently, many methods that should not be user facing are user facing:
- `init_morph()`
- `get_states_from_nodes_and_edges()`
- `initialized`
- `initialize()`

Possibly also:
- `to_jax()`
- `get_all_parameters()`

but those might be needed once we expose the update loop.

Also, `Compartment.init_conds()`, `Cell.update_summed_coupling_conductances()` (why do we still have this at all?), `Network.init_syns()`.

michaeldeistler:
Closes #448

### DIFF
--- a/jaxley/modules/network.py
+++ b/jaxley/modules/network.py
@@ -105,8 +105,8 @@ class Network(Module):
         # Channels.
         self._gather_channels_from_constituents(cells)
 
-        self.initialize()
-        self.init_syns()
+        self._initialize()
+        self._init_syns()
 
     def __repr__(self):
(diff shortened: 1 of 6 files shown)


jaxley/modules/network.py
```diff
@@ -105,8 +105,8 @@ def __init__(
         # Channels.
         self._gather_channels_from_constituents(cells)
 
-        self.initialize()
-        self.init_syns()
+        self._initialize()
+        self._init_syns()
```
jnsbck, line 109: Did you also change `_init_syns` in the definition of `Network`?

jnsbck (approved):
I assume the methods you removed are obsolete.
What about `init_states` ?
Also please check the comment I left. otherwise lgtm.

PR #461, opened 2024-10-23
fix units of compute_current

### LINKED ISSUE #274: Units of `compute_current()` in channels do not match NEURON
- Conversion to `uA/cm^2` of external input is [here](https://github.com/jaxleyverse/jaxley/blob/7311cd70e04eb87402041b816ddec498b323ed6d/jaxley/modules/base.py#L742)
- `compute_current()` of channels also returns `mS/cm^2 * mV`, which is `uA/cm^2`, see [here](https://github.com/jaxleyverse/jaxley/blob/7311cd70e04eb87402041b816ddec498b323ed6d/jaxley/channels/hh.py#L51-L60)
- As written [here](https://www.neuron.yale.edu/neuron/static/docs/units/unitchart.html) the capacitance in `um/cm2`.

In summary: all is correct, but (unlike is written in the documentation), the `compute_current()` method of a `Channel` returns the current in `uA/cm2`, **not** in `mA/cm2`.

michaeldeistler:
Closes #274 

This is a tricky PR as it will impact users which ported channels themselves. I hope that this is not many people yet, so now is the time to still do it.


huangziwei (changes requested):
I would simply get rid of the backward compatibility and force users to use the new unit convention instead. Having too many options also makes it difficult to debug...

huangziwei:
I am a bit confused by all the unit conversions now... but do we also need to change `convert_point_process_to_distributed`? (as it's converting the injected current in nA to uA/cm2?)

Also, the units in the Channel class' doc string also needs to be fixed .

huangziwei (approved):
Looks good!

Issue #350, opened 2024-04-23
Expose `step` such that users have fine-grained control over solve


michaeldeistler:
If one wants to change, for example, `dt` during training one currently has to recompile. Instead, it would be nice to just compile `step` once and then let the user write the update loop.

manuelgloeckler:
This is also nice for some inference algorithms like particle filters that only require the step function.

PR #466, opened 2024-10-24
User exposing init and step for customization.

manuelgloeckler:
# User exposing init and step

Currently, the only user API to solve a cell ODE is `jx.integrate`. For various applications, it is useful to have access to an `init` function, which initializes the state, and a `step` function, which executes a single ode step. Some examples are:
- Nonuniform step sizes
- Integration with Kalman or Particle filtering frameworks
- Stochastic Models (i.e. add noise after each step) ...

Also see #350 .

## Proposal

```python
init_fn, step_fn = build_init_and_step_fn(cell, **maybe_some_other)

# All the reformating
states, params = init_fn(params, all_states, **kwargs)
next_state = step_fn(states, params, externals, dt=0.025, **kwargs)
```

### DIFF
--- a/jaxley/integrate.py
+++ b/jaxley/integrate.py
@@ -2,7 +2,7 @@
 # licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
 
 from math import prod
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import jax.numpy as jnp
 import pandas as pd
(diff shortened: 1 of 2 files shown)


michaeldeistler (commented):
Beautiful, I love it!

Could you write a test (which we can later reuse as a tutorial :) ) which writes a compares `jx.integrate()` to a minimal version of iteratively calling `step()`?

manuelgloeckler:
Alright, I added an API equivalence test of integrate vs. manual for-loop.

michaeldeistler (approved):
Amazing, thanks!

PR #470, opened 2024-10-24
Write trainables to module

### LINKED ISSUE #356: update `Module` after training
Is there a way to update the parameters of `Module` after training, i.e. something like `from_jax`? I think this would be good great to have. For example if you want to look at distribution of updated radii, lengths, params with `vis`.

michaeldeistler:
This PR introduces a `write_trainables()` which adds the trainables to the `.nodes` and `.edges` such that they can be used for plotting, printing,...

```python
net.make_trainable("radius")
params = net.get_parameters()

net.write_trainables(params)
```

Closes #356

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -682,8 +682,8 @@ class Module(ABC):
         self.base.jaxedges = {}
         edges = self.base.edges.to_dict(orient="list")
         for i, synapse in enumerate(self.base.synapses):
+            condition = np.asarray(edges["type_ind"]) == i
             for key in synapse.synapse_params:
-                condition = np.asarray(edges["type_ind"]) == i
                 self.base.jaxedges[key] = jnp.asarray(np.asarray(edges[key])[condition])
             for key in synapse.synapse_states:
                 self.base.jaxedges[key] = jnp.asarray(np.asarray(edges[key])[condition])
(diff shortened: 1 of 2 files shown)


jaxley/modules/base.py
```diff
@@ -1044,6 +1045,32 @@ def make_trainable(
                 f"Number of newly added trainable parameters: {num_created_parameters}. Total number of trainable parameters: {self.base.num_trainable_params}"
             )
 
+    def write_trainables(self, trainable_params: List[Dict[str, jnp.ndarray]]):
```
jnsbck, line 1048: Cool! This only works for Module tho, i.e. doing `net.cell(0).write_trainables()` will only write to `self.base.nodes` if `self == self.base`. To be able to call this from a `View` and for these changes to take affect for the module you need to write to `self.base.nodes.loc[self._nodes_in_view]`, instead of `self.nodes`. See docstring of `View`. I think it might actually be nice if you could write only the trainables in View.

tests/test_make_trainable.py
```diff
@@ -421,3 +421,40 @@ def simulate(params):
     parameters = net.get_parameters()
     v = simulate(parameters)
     assert np.invert(np.any(np.isnan(v))), "Found NaN in voltage."
+
+
+def test_write_trainables():
```
jnsbck, line 426: Can you also add a check to ensure that the params in nodes have changed?

jnsbck (commented):
Cool!, nice that it needs so few lines of code!

jaxley/modules/base.py
michaeldeistler: doesn't work, see comments

jnsbck (approved):
Feel free to merge this.
Regarding the special treatment of edges: since we are getting rid of synapse view down the line, we can treat edges like nodes. I think this would make a lot of things nicer. Currently working on this as part of #453.

PR #475, opened 2024-10-29
Add runtime tests

michaeldeistler:
Note that the tests are disabled in the workflow: `pytest -m "not run_locally"`


.github/workflows/regression_tests.yml
```diff
@@ -0,0 +1,60 @@
+name: Regression Tests
```
jnsbck, line 1: I tested this out in a seperate repo, but I cannot for the life of me get the commenting to work.
The basic idea is:
1. run regression tests -> output to file `regression_test_results.json`
2. get the diff between `regression_test_results.json` and `main:regression_test_results.json`
3. comment on PR with sth. along the lines of "test x is slower than in main"
4. ammend last commit, such that when PR gets merge `regression_test_results.json` becomes new baseline.

We could also do without the ammend, then we would have to maintain the baseline file manually. This might be preferred actually since it gives us full control over the baseline.

tests/regression_test_runner.py
```diff
@@ -0,0 +1,75 @@
+# tests/benchmark_runner.py
```
jnsbck, line 1: used to run the benchmark and generate the report

tests/test_regression.py
```diff
@@ -0,0 +1,14 @@
+# your_module.py
```
jnsbck, line 1: This is just an example, would essentially put the functions that you have written here.

tests/test_runtime.py
```diff
@@ -0,0 +1,124 @@
...
+import pytest
+from jax import jit
+
+import jaxley as jx
+from jaxley.channels import HH
+from jaxley.connect import sparse_connect
+from jaxley.synapses import IonotropicSynapse
+
+
+def build_net(num_cells, artificial=True, connect=True, connection_prob=0.0):
```
jnsbck, line 17: Would it make sense to also time building the net and connecting?

tests/test_runtime.py
```diff
@@ -0,0 +1,124 @@
...
+        5: 13.303060293197632,
+    }
+    run_times = {
+        0: 0.08291006088256836,
+        1: 0.596994161605835,
+        2: 0.8518729209899902,
+        3: 5.746302127838135,
+        4: 1.3585789203643799,
+        5: 12.48916506767273,
+    }
```
jnsbck, line 120: The problem with these hardcoded times is that they are system dependent. Did you get these from the github actions runner?

jnsbck (commented):
Awesome! Thanks for implementing this, I think this will be really useful. I had also worked on the same thing, but was pursuing sth slightly different. I have added my proof of concept to this branch. (see also my comments). Feel free to work off of it or remove it. I think having a seperate CI for regressions might be useful in the long run.

Happy to discuss this more.

jnsbck:
I have now added the following:
- `tests/test_regression.py`: Took your runtime tests and changed the structure, such that they can be run with my draft CI.
- `tests/regression_test_runner.py`: Runs the test functions in `test_regression.py`, times them, saves the results and generates a report.
- `regression_tests.yml`: This action runs the `regression_test_runner.py` and comments on the PR with the results of the tests. Rather than commenting we could also make this a requirement for the action to pass, i.e. `return 1` from the runner instead of generating the report. 
- `update_regression_baseline.yml`: This can be used to update the regression tests in main, should this be desired. This has to be run manually and adds the new baselines to the PR. (this is not tested and wip).

Lemme know your thoughts and if this makes sense.

jnsbck:
I invested way too much time into this, but I basically scrapped the prev idea from [5d76612](https://github.com/jaxleyverse/jaxley/pull/475/commits/5d76612ee3def9e2eac69696b8d3f9351c24b91a) and redid the whole thing. Now its integrated into pytest (more similar than what you had proposed before) and hence should be easier to maintain.

I added the following:
- github action to update the baseline: This should be triggered manually.
- github action to run the regression tests: checks out the baselines from the main branch, records the new runtimes and compares them against the appropriate baseline.
- more sophisticated `test_regression.py`: Includes a bunch of helpful things to make running, storing and comparing the results easier.

jnsbck (commented):
-

.github/workflows/regression_tests.yml
```diff
@@ -0,0 +1,39 @@
...
+    
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+        architecture: 'x64'
+    
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e ".[dev]"
```
michaeldeistler, line 28: do we really need dev?

.github/workflows/update_regression_baseline.yml
```diff
@@ -0,0 +1,88 @@
...
+    
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+        architecture: 'x64'
+    
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e ".[dev]"
```
michaeldeistler, line 44: dev needed?

.github/workflows/update_regression_baseline.yml
```diff
@@ -0,0 +1,88 @@
+# .github/workflows/update_regression_tests.yml
+
+# for details on triggering a workflow from a comment, see:
+# https://dev.to/zirkelc/trigger-github-workflow-for-comment-on-pull-request-45l2
+name: Update Regression Baseline
+
+on:
+  issue_comment: # trigger from comment; event runs on the default branch
+    types: [created]
+  # pull_request:
```
michaeldeistler, line 10: delete?

.github/workflows/update_regression_baseline.yml
```diff
@@ -0,0 +1,88 @@
...
+          });
+                
+    - name: Commit and push
+      if: github.event.pull_request.base.ref == 'main'
+      run: |
+        git add -f tests/regression_test_baselines.json # since it's in .gitignore
+        git commit -m "Update regression test baselines"
+        git push origin HEAD:${{ github.head_ref }}
+    
+    # Needed when workflow is triggered from a comment
```
michaeldeistler, line 82: delete?

.gitignore
```diff
@@ -55,6 +55,8 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+tests/regression_test_results.json
+tests/regression_test_baselines.json
```
michaeldeistler, line 59: I don't get this. Why do we first put it in the gitignore just to then force add it?

tests/test_regression.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # Save data to a JSON file
+    result_data = load_json(fpath)
+    result_data.update(data)
+
+    with open(fpath, "w") as f:
+        json.dump(result_data, f, indent=2)
+
+
+class compare_to_baseline:
```
michaeldeistler, line 126: `CompareToBaseline` to follow PEP style

michaeldeistler (commented):
Wow this is amazing!!!

I don't think I understand when exactly the updating of the times is being run. On a comment to this particular PR (i.e., #475)? Or some other PR/issue? Also, what should the comment be? Anything? Who would the comment have to be by? Anyone? (BTW I am commenting now, so does this trigger the regression tests :D ?)

.github/workflows/regression_tests.yml
jnsbck: nope, good catch

jnsbck:
The event `issue_comment` gets triggered **on every** comment that gets posted to an issue or PR. However, the workflow to produce an updated baseline is only triggered if the comment body contains the string `/update_regression_baseline` (so your comment would not have triggered it a) because the `issue_comment` workflow has to be merged to main first due to the way the event listener works on github and b) cos you did not use the command above). The workflow is run on the PR branch, which means the new baseline is produced from the updated codebase. The change to the baselines is then merged into the current PR. The command to start the workflow can currently be used by all users, but we could limit this to admins as well.

.gitignore
jnsbck: I agree this appears a bit odd, but I want to prevent users from pushing an updated baseline file from their local machine (which is why I have added it to the gitignore). The only way baselines should be updated is through github actions (to ensure regression tests are run in a consistent environment). That is why the github actions uses -f, while for the user changes to these files are ignored.

jnsbck: Does that make sense?

jnsbck:
I have verified this to work in a seperate repo, if the workflow to update the baselines exists on the default branch. Upon commenting with `/update_regression_baselines` on any PR, the workflow gets triggered. First a comment is posted to inform the user the workflow is run, then the new baselines are computed and compared to the ones stored in main, and a report is produced and the prev. comment is updated with the result. If successful, the updated baselines are pushed to the PR branch if the branch is part of the repo or a seperate PR is opened in case the branch is from a fork (The user is also made aware of this in the comment), see below.

image

Since the workflow can only be triggered on the default branch this PR has to be merged first and needs a seperate PR to update the baselines using `/update_regression_baselines`. Hence the regression tests above failing. 

Side note: The report is always written to the pytest output (above the Failure report), so you can check why the regression test action might have failed in detail.

image

With this, I think this can be merged @michaeldeistler, unless you have any other reseverations.

.gitignore
michaeldeistler: Ah, got it, yes makes sense!

michaeldeistler:
Thanks a lot! Feel free to merge (I cannot approve because it is technically "my" PR).

Issue #449, opened 2024-10-17
Refactor tests


jnsbck:
Tidy up structure of `tests/`.
Add [test fixtures](https://docs.pytest.org/en/stable/how-to/fixtures.html) for some of the repeatedly used modules, i.e.

```python
import pytest

@pytest.fixture
def fully_connected_net():
    # build net and connect it
    # add mechanisms etc.
    return net


def some_test():
    assert fully_connected_net.cell(0)....
```

Add sensible [markers](https://docs.pytest.org/en/stable/how-to/mark.html) to tests, i.e. `slow` or other things.

fabioseel:
For reference, this is the runtimes of the individual tests:

```
38.77s call     tests/test_plotting_api.py::test_volume_plotting
32.74s call     tests/test_swc.py::test_swc_voltages[morph.swc]
31.66s call     tests/test_swc.py::test_swc_voltages[morph_single_point_soma.swc]
22.57s call     tests/jaxley_identical/test_swc.py::test_swc_cell[morph_single_point_soma.swc-jaxley.stone]
22.24s call     tests/jaxley_identical/test_swc.py::test_swc_cell[morph.swc-jaxley.stone]
19.03s call     tests/jaxley_identical/test_grad.py::test_network_grad
15.12s call     tests/test_cell_matches_branch.py::test_equivalence
13.45s call     tests/test_optimize.py::test_type_optimizer
11.75s call     tests/jaxley_identical/test_swc.py::test_swc_net[jaxley.stone]
7.70s call     tests/test_grad.py::test_branch_grad_against_finite_diff_initial_state[HH_m]
7.07s call     tests/test_channels.py::test_init_states
6.17s call     tests/test_api_equivalence.py::test_solver_backends_net
5.82s call     tests/test_set_ncomp.py::test_api_equivalence_swc_lengths_and_radiuses[morph_250.swc-5]
5.76s call     tests/jaxley_identical/test_basic_modules.py::test_complex_net[jaxley.stone]
5.37s call     tests/test_make_trainable.py::test_data_set_vs_make_trainable_network
5.27s call     tests/test_set_ncomp.py::test_api_equivalence_swc_lengths_and_radiuses[morph_250.swc-1]
5.23s call     tests/test_grad.py::test_branch_grad_against_finite_diff_initial_state[v]
5.01s call     tests/test_make_trainable.py::test_data_set_vs_make_trainable_pospischil
[...]

PR #479, opened 2024-11-04
Speed up tests (swc, plotting)

### LINKED ISSUE #468: Speed up plotting (reduce mesh resolution)
The level of detail of the meshes (create_cone_frustum_mesh, create_cylinder_mesh, create_sphere_mesh) is the main bottleneck for plotting faster (for morph plots).
However, all three of this methods start with a line defining a 'resolution'. I think it would be nice to make this a parameter and allow to produce low-resolution but fast plots.
Also useful to have the plot tests run faster ;)

fabioseel:
Main contributions:
- swc tests now significantly faster (change to sparse solver)
- plotting tests a lot faster
  - exposed 'resolution' of mesh generation (no proper exposition to user (yet)) fixes #468
  - splitted plot volume test for more granular analysis

minor contributions:
- added test fixtures for compartment, cell, morph_cell, branch and network -> not sure how much it helps / improves readability here. Could also revert these changes. (might be a first step for #449)

### DIFF
--- a/tests/test_plotting_api.py
+++ b/tests/test_plotting_api.py
@@ -172,21 +172,21 @@ def test_mixed_network(morph_cell, cell):
 def test_volume_plotting_2d(comp, branch, cell, simple_net, morph_cell):
     fig, ax = plt.subplots()
     for module in [comp, branch, cell, simple_net, morph_cell]:
-        module.vis(type="comp", ax=ax)
+        module.vis(type="comp", ax=ax, morph_plot_kwargs={"resolution": 6})
     plt.close(fig)
 
 
 def test_volume_plotting_3d(comp, branch, cell, simple_net, morph_cell):
     # test 3D plotting
     for module in [comp, branch, cell, simple_net, morph_cell]:
-        module.vis(type="comp", dims=[0, 1, 2])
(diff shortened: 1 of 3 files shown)


tests/test_plotting_api.py
```diff
@@ -19,31 +19,58 @@
 from jaxley.synapses import IonotropicSynapse
 
 
-def test_cell():
-    dirname = os.path.dirname(__file__)
-    fname = os.path.join(dirname, "swc_files", "morph.swc")
-    cell = jx.read_swc(fname, nseg=4)
+@pytest.fixture(scope="module")
```
jnsbck, line 22: This is awesome! I think we could even move them to a separate file, to make them available to all other tests. `tests/fixtures.py`.

jnsbck: Feel free to replace as many of the other tests that rely on `comp` or `net` etc. with these fixtures. But also fine if you don't and then just leave #449 open and just add a quick comment to it.

jnsbck (approved):
Not a lot to critique here. Left only a couple comments. Feel free to merge this once adressed. :)

michaeldeistler (approved):
Awesome! Good to go also from my side! @fabioseel let us know if/when we can merge it! Thanks a ton!

jnsbck:
Awesome, thanks a ton!

PR #480, opened 2024-11-04
Add CI to check if tutorials can be run.

jnsbck:
This PR adds a github action to run all tutorials.

Q: what should the trigger for this be? Not sure if we should run this on every push/pull request.

Might be sufficient if we run this when a PR is marked with `prepare_release` or something.


jnsbck:
- [ ] rm tensorflow from `.toml` once #484 is merged

michaeldeistler (approved):
Awesome!

From my perspective it is fine as it is---always run it once we push to main (which is also when the website is hosted). Feel free to merge!

PR #484, opened 2024-11-04
Remove tensorflow dependency.

### LINKED ISSUE #481: `replace tensorflow.Dataset` by `optimize.utils.Dataset`
We do not list `tensorflow` as a dependency, but you need it to run tutorial 07. Rather than adding `tensorflow` to the dev dependencies, I'd suggest writing a small `Dataset` class for this case. Also relevant to https://github.com/jaxleyverse/jaxley/pull/480.

https://github.com/jaxleyverse/jaxley/blob/981ab6412e8894ada36be346dddb17ec65e7ab05/docs/tutorials/07_gradient_descent.ipynb#L759

jnsbck:
Added small custom dataloader tutorial 07 to get rid of tensorflow dependency. this is the only tensorflow dependency that we have. 

Closes #481.


jnsbck:
@michaeldeistler  fyi

michaeldeistler:
Thanks!

PR #487, opened 2024-11-07
Simpler edge indexing

jnsbck:
Currently synapse and channel parameters / states are handled differently. While channel params are referred to with global indices,  synapse params are referenced on a per synapse basis. This leads to very different implementations of synapse and channel updates. This PR makes the synapse indexing global, which simplifies several aspects of the code.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -32,7 +32,7 @@ from jaxley.utils.cell_utils import (
     interpolate_xyz,
     loc_of_index,
     params_to_pstate,
-    query_channel_states_and_params,
+    query_states_and_params,
     v_interp,
 )
 from jaxley.utils.debug_solver import compute_morphology_indices
@@ -712,16 +712,15 @@ class Module(ABC):
         """
         jaxnodes = self.base.jaxnodes = {}
         nodes = self.base.nodes.to_dict(orient="list")
-        
+
         jaxedges = self.base.jaxedges = {}
         edges = self.base.edges.to_dict(orient="list")
-        edges.pop("type") # drop since column type is string
-        
+        edges.pop("type")  # drop since column type is string
+
         for jax_array, params in zip([jaxnodes, jaxedges], [nodes, edges]):
             for key, value in params.items():
                 inds = jnp.arange(len(value))
                 jax_array[key] = jnp.asarray(value)[inds]
-
 
     def show(
         self,
@@ -1230,17 +1229,6 @@ class Module(ABC):
             inds = parameter["indices"]
             set_param = parameter["val"]
 
-            # This is needed since SynapseViews worked differently before.
-            # This mimics the old behaviour and tranformes the new indices
-            # to the old indices.
-            # TODO FROM #447: Longterm this should be gotten rid of.
-            # Instead edges should work similar to nodes (would also allow for
-            # param sharing).
-            synapse_inds = self.base.edges.groupby("type").rank()["global_edge_index"]
-            synapse_inds = (synapse_inds.astype(int) - 1).to_numpy()
-            if key in self.base.synapse_param_names:
-                inds = synapse_inds[inds]
-
             if key in params:  # Only parameters, not initial states.
                 # `inds` is of shape `(num_params, num_comps_per_param)`.
                 # `set_param` is of shape `(num_params,)`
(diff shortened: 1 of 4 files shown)


jaxley/modules/network.py
```diff
@@ -261,40 +262,29 @@ def _step_synapse_state(
...
-        post_syn_inds = grouped_syns["global_post_comp_index"].apply(list)
-        synapse_names = list(grouped_syns.indices.keys())
+        for i, group in edges.groupby("type_ind"):
+            synapse = syn_channels[i]
+            pre_inds = group["global_pre_comp_index"].to_numpy()
+            post_inds = group["global_post_comp_index"].to_numpy()
+            edge_inds = group.index.to_numpy()
 
-        for i, synapse_type in enumerate(syn_channels):
-            assert (
```
jnsbck, line 270: Is this still relevant?

jnsbck:
- [ ] rerun tests after #492 is merged

michaeldeistler (commented):
Nice!

I am a bit worried about memory consumption though. If every `jaxedge` has every state, then there will a lot of NaNs each of which contains memory. I think that, in the worst case, memory increases by N**2, where is the number of types of synapses.

Can we have a zoom about this? E.g. next week?

jnsbck:
True! I did not think about this. I have two thoughts though. Why not do this for nodes on a per mechanism basis as well. And is there not a more straight forward way to avoid nans while still using global indexes? 
But yes, maybe a quick zoom would he great :)

michaeldeistler:
I agree that we should also do it like this for nodes. Let's zoom next week!

jaxley/utils/cell_utils.py
```diff
@@ -774,3 +759,16 @@ def compute_children_and_parents(
...
+    concat_df = pd.concat(dfs, ignore_index=True)
+    # replace nans with Nones
+    # this correctly casts float(None) -> NaN, bool(None) -> NaN, etc.
+    concat_df[concat_df.isna()] = None
+    for col in concat_df.columns[concat_df.dtypes == "object"]:
+        for df in dfs:
+            if col in df.columns:
+                concat_df[col] = concat_df[col].astype(df[col].dtype)
+            break  # first match is sufficient
+    return concat_df
```
jnsbck, line 774: fixes a bug, where dtype of bool columns was changed upon concatenation in:
```python
comp1 = jx.Compartment()
comp2 = jx.Compartment()
comp2.insert(HH())
branch = jx.Branch([comp1, comp2)] #-> branch.nodes["HH"].dtype was `object` in this case
```

jaxley/modules/base.py
```diff
@@ -189,7 +185,9 @@ def __str__(self):
 
     def __dir__(self):
         base_dir = object.__dir__(self)
-        return sorted(base_dir + self.synapse_names + list(self.group_nodes.keys()))
```
jnsbck, line 192: `group_nodes` was removed in #447.

jnsbck:
Main changes:
- rename `channel_params`... `synapse_states` to `params`...`states` (allows to reuse stuff for both synapses and channels)
- `to_jax` stores params/states on a per mechanism basis now -> no more NaNs. To do this some prep is needed in order to keep track of what goes where to find it again later. Hence I had to write a `_prepare_for_jax` method, that pre-computes indices and adds a lookup method for all states / params in the module. This should also make this easily adaptable if we add the global_param API change.
- refactor of `get_all_parameters` and `get_all_states`
- refactor of channel and synapse updates.

In general I think the code is a bit cleaner and more consolidated now (-150 line diff). Will also run regression tests. Maybe its even faster as well. I am not super happy about the `_prepare_for_jax` solution tbh, but it seems to work well fwiw. If you have feedback, I'd be very happy about any suggestions!

Lemme know what you think or if you have any questions.

The only thing I touched in terms of channels / synapse rewrite is the params/states rename. Does this warrant a seperate branch?

michaeldeistler (approved):
Thanks a lot @jnsbck!

I am already off for the holidays so will not be able to do a detailed review, but:
1) I think all your high-level explanations sound good!
2) Yes, let's make a new `v1.0` branch and merge it there because it breaks all channel models.

And yes, please run regression tests and ensure that all is good! Good to go then! Thanks!

jnsbck:
I have done some refactoring and am now much happier with how the code looks. Locally tests pass, but since this is not a pull request to main (but v1.0), tests are not being run here. I also started another regression tests (see above), which I hope will be as quick as prev, i.e. faster than whats in `main`.

Would still be great if you could have a more thorough look before I merge it, when your back in the office.
Also, this PR should work with shared states for synapses as well now, which we did not work prior and which we do not check. Lemme know if I should add a test

jnsbck:
...more refactoring and consolidation imo. Somewhat forward thinking, now `to_jax` moves the `jx.Module` to a single `pytree`, and makes handling global params and states much simpler now. Regression test should also pass again.

jnsbck:
...more refactoring and consolidation imo. Somewhat forward thinking, now `to_jax` moves the `jx.Module` to a single `pytree`, and makes handling global params and states much simpler now. Regression test should also pass again. 

/test_regression

michaeldeistler (approved):
Thanks, LGTM!

PR #489, opened 2024-11-07
Connectivity_matrix_connect update

### LINKED ISSUE #486: Bug in connectivity_matrix_connect
The following code does not work because of issues indexing into the precell indices when creating the rows for appending multiple synapses; it only works in the pre range starts at zero.

```
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=1)
cell = jx.Cell([branch], parents=[-1])
net = jx.Network([cell for _ in range(10)])

pre = net.cell(list(range(2, 5)))
post = net.cell(list(range(5, 7)))
print(len(pre.nodes), len(post.nodes))

conn_mat = np.ones((len(pre.nodes), len(post.nodes))).astype(bool)
print(conn_mat)

connectivity_matrix_connect(pre, post, IonotropicSynapse(), conn_mat)
```
One gets an index error
image

NOTE: this worked about a month ago but not anymore

I don't think this was intended right @jnsbck?

kyralianaka:
This PR resolves #486 so that views with global indices of different ranges can be used to connect cells with an adjacency matrix (and adds a test). 

I also removed the random selection of compartments on the post synaptic cell because I don't quite see the benefit to doing this as opposed to just using the first compartment of the first branch, and the random selection in list comprehension was really slowing down my building of large networks. This being said, I would be interested in having connectivity_matrix_connect take a matrix (num pre comps x num post comps) instead of (num pre cells x num post cells) for more fine control of the connectivity... I can open another issue about this as enhancement.

### DIFF
--- a/jaxley/connect.py
+++ b/jaxley/connect.py
@@ -134,7 +134,8 @@ def connectivity_matrix_connect(
 
     Connects pre- and postsynaptic cells according to a custom connectivity matrix.
     Entries > 0 in the matrix indicate a connection between the corresponding cells.
-    Connections are from branch 0 location 0 to a randomly chosen branch and loc.
+    Connections are from branch 0 location 0 on the presynaptic cell to branch 0
+    location 0 on the postsynaptic cell.
 
     Args:
         pre_cell_view: View of the presynaptic cell.
@@ -142,7 +143,7 @@ def connectivity_matrix_connect(
         synapse_type: The synapse to append.
         connectivity_matrix: A boolean matrix indicating the connections between cells.
     """
-    # Get pre- and postsynaptic cell indices.
+    # Get pre- and postsynaptic cell indices
     pre_cell_inds = pre_cell_view._cells_in_view
     post_cell_inds = post_cell_view._cells_in_view
 
@@ -152,24 +153,19 @@ def connectivity_matrix_connect(
     ), "Connectivity matrix must have shape (num_pre, num_post)."
     assert connectivity_matrix.dtype == bool, "Connectivity matrix must be boolean."
 
-    # get connection pairs from connectivity matrix
+    # Get connection pairs from connectivity matrix
     from_idx, to_idx = np.where(connectivity_matrix)
-    pre_cell_inds = pre_cell_inds[from_idx]
-    post_cell_inds = post_cell_inds[to_idx]
 
-    # Sample random postsynaptic compartments (global comp indices).
-    global_post_indices = np.hstack(
-        [
-            sample_comp(post_cell_view.scope("global").cell(cell_idx))
-            for cell_idx in post_cell_inds
-        ]
-    )
-    post_rows = post_cell_view.nodes.loc[global_post_indices]
-
-    # Pre-synapse is at the zero-eth branch and zero-eth compartment.
+    # Pre-synapse at the zero-eth branch and zero-eth compartment
     global_pre_indices = (
         pre_cell_view.scope("local").branch(0).comp(0).nodes.index.to_numpy()
     )  # setting scope ensure that this works indep of current scope
-    pre_rows = pre_cell_view.select(nodes=global_pre_indices[pre_cell_inds]).nodes
+    pre_rows = pre_cell_view.select(nodes=global_pre_indices[from_idx]).nodes
+
+    # Post-synapse also at the zero-eth branch and zero-eth compartment
+    global_post_indices = (
+        post_cell_view.scope("local").branch(0).comp(0).nodes.index.to_numpy()
+    )
+    post_rows = post_cell_view.select(nodes=global_post_indices[to_idx]).nodes
 
     pre_cell_view.base._append_multiple_synapses(pre_rows, post_rows, synapse_type)
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -203,3 +203,17 @@ def test_connectivity_matrix_connect():
     comp_inds = nodes.loc[net.edges[cols].to_numpy().flatten()]
     cell_inds = comp_inds["global_cell_index"].to_numpy().reshape(-1, 2)
     assert np.all(cell_inds == incides_of_connected_cells)
+
+    net = jx.Network([cell for _ in range(4 * 4)])
+    connectivity_matrix_connect(
+        net[1:4], net[2:6], TestSynapse(), m_by_n_adjacency_matrix
+    )
+    assert len(net.edges.index) == 5
+    nodes = net.nodes.set_index("global_comp_index")
+    cols = ["global_pre_comp_index", "global_post_comp_index"]
+    comp_inds = nodes.loc[net.edges[cols].to_numpy().flatten()]
+    cell_inds = comp_inds["global_cell_index"].to_numpy().reshape(-1, 2)
+    # adjust the cell indices based on the view range passed
+    incides_of_connected_cells[:, 0] += 1
+    incides_of_connected_cells[:, 1] += 2
+    assert np.all(cell_inds == incides_of_connected_cells)


jnsbck:
This was also fixed by #488, which I merged already. 
The random selection is consistent with the other cell level connection methods. Also the first of the first of the first will not work with single comp cells. I agree that we should make it faster though if it slows down the whole thing.

The comp2comp connectivity matrix connect that you're proposing is basically already possible using `connect` since it will not just take single comps but also n comps. Something like the following should work:

```python
con_matrix = some_boolean_matrix
pre_inds, post_inds = np.where(con_matrix)
pre_comps = net.select(pre_inds)
post_comps = net.select(post_inds)

connect(pre_comps, post_comps, TestSynapse())
```

kyralianaka:
I am happy to go through the other methods and remove the random selection; it will always make connecting networks faster and avoid having to look for the postsynaptic compartments.

I'm not sure why you say it won't work for single compartment cells, I can change the test to single compartment cells and have no problem. 

For the comp2comp mat, that's fine, no need to add then. 

In general I would just like to make this function easier to read, mainly by not using so many types of indices for the same cells ("from_idx", "pre_cell_inds", and "global_pre_cell_inds"). I think there are even more simplifications I can do, so I will do those and commit again (will also remove the rand selection).

kyralianaka:
I also think the added test is really important especially if you make more plans to change synapse indexing

jnsbck:
I agree that it will make networks faster, but I would not make it the default. I'd be fine with making it a toggle though, i.e.
`post_comp="random"` per default and sth. like `post_comp=1` for branch 0, comp 1. If we add this, we should keep in mind that branches can have different numbers of segments tho.

Ah I see, you mean 0th comp rather than 1st. In that case I am strongly against hard coding this, since we would always connect soma to soma.

I'd be surprised if you can make rewrite it without `"from_idx", "pre_cell_inds", and "global_pre_cell_inds", but if you have an idea how if can work its great if you can make the function easier to read.

Could you briefly explain what the additional test is covering that is not covered by the other cases? Also please add some more doc/comments :)

kyralianaka:
I am strongly for making it the default haha, but open to giving the option for random selection. Your concern about always connecting soma to soma doesn't quite make sense to me because the soma doesn't have to be and often isn't the zeroith compartment (when self-defining cells, I could be wrong about swc loaded cells, in which case I would reconsider). 

Re all the indices, I renamed things and I think the most clarifying aspect is distinguishing between global_cell_inds and global_comp_inds. I also really like it if the pre and post indices are handled the same way because this makes the function more modular. 

I just added some comments to the tests, they are in the test_connectivity_matrix_connect() function, but you can let me know if anything is still unclear. I might still go add even more comments :)

kyralianaka:
Thinking about the way cells are typically constructed, would it actually maybe make sense to default connect the last compartment of the presynaptic cell to the zeroith compartment of the post-synaptic cell? (yes before I meant zeroith when I said first sorry for the confusion)

kyralianaka:
Another thing, the test_sparse_connect() function uses this:
` assert all(
        [
            63,
            59,
            65,
            86,
            80,
            58,
            92,
            85,
            168,
            145,
            189,
            153,
            180,
            190,
            184,
            163,
            159,
            179,
            182,
        ]
    )`
so it will always pass... I will fix that here

kyralianaka:
I updated all of the connectivity functions now so that they all index in the same way. It probably looks more repetitive now, but I would argue that that's a good thing as a more modular design to avoid potential future bugs. 

I haven't yet included the option to randomly select the postsynaptic compartment, but I guess we should talk about that as a group and get everyone's opinions.

kyralianaka:
Two tests are currently failing because changing the post_comp_index of the fully connected network changes the voltage solutions of simulations. I will wait to fix this then until we talk about the post_comp_index scheme.

kyralianaka:
I just added back the option to randomly select the post-synaptic compartment; setting this to true makes most of the old tests pass (except where I changed other things). 

The other thing I changed was the way of sparsely connecting the network because the previous method randomly sampled pre and post cells separately, which resulted in a lot of duplicate connections. Now the connectivity matrix is generated by independent Bernoulli trials, consistent with common random graph generation methods. The one disadvantage of this method is that it could be memory intensive for extremely large numbers of pre and post synaptic cells, but since the matrix is local, I'm hopeful that it won't cause problems. 

I might look into the (comp, comp) matrix situation in another pull request since it will soon become relevant to my project, but hopefully we can update these functions first :)

michaeldeistler:
Cool! When you are done please request a review!

tests/test_connection.py
```diff
@@ -203,3 +208,38 @@ def test_connectivity_matrix_connect():
     comp_inds = nodes.loc[net.edges[cols].to_numpy().flatten()]
     cell_inds = comp_inds["global_cell_index"].to_numpy().reshape(-1, 2)
     assert np.all(cell_inds == incides_of_connected_cells)
+
+    # Test with different view ranges
```
michaeldeistler, line 212: what is a `view range`?

tests/jaxley_identical/test_grad.py
```diff
@@ -33,13 +33,13 @@ def test_network_grad():
     _ = np.random.seed(0)
     pre = net.cell([0, 1, 2])
     post = net.cell([3, 4, 5])
-    fully_connect(pre, post, IonotropicSynapse())
-    fully_connect(pre, post, TestSynapse())
+    fully_connect(pre, post, IonotropicSynapse(), random_post_comp=True)
+    fully_connect(pre, post, TestSynapse(), random_post_comp=True)
```
michaeldeistler, line 37: is there a test for `random_post_comp=False`? At least for `fully_connect`, but also for the others? At least to check that they do not break?

jaxley/connect.py
```diff
@@ -129,49 +149,54 @@ def connectivity_matrix_connect(
...
     synapse_type: "Synapse",
     connectivity_matrix: np.ndarray[bool],
+    random_post_comp: bool = False,
 ):
     """Appends multiple connections which build a custom connected network.
 
     Connects pre- and postsynaptic cells according to a custom connectivity matrix.
     Entries > 0 in the matrix indicate a connection between the corresponding cells.
-    Connections are from branch 0 location 0 to a randomly chosen branch and loc.
+    Connections are from branch 0 location 0 on the presynaptic cell to branch 0
```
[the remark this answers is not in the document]
michaeldeistler: at least add `by default` and mention `random_post_comp`

michaeldeistler (changes requested):
Awesome, thanks! Could you report the runtime of `random_post_comp = True` in the PR description and compare it to the runtime before this PR? Especially for large networks.

(And ideally also add the timing for `random_post_comp=False` (after this PR).

tests/test_connection.py
```diff
@@ -84,41 +84,42 @@ def test_fully_connect():
 
     fully_connect(net[8:12], net[12:16], TestSynapse())
 
+    # This was previously visually inspected
```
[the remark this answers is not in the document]
kyralianaka: Here I just mean that I ran network.vis() in a notebook many times to make sure that the connectivity looks as expected, but I can remove these comments

tests/jaxley_identical/test_grad.py
kyralianaka: All of the tests in test_connection.py are with random_post_comp=False, but the tests that used connection.py everywhere else (test_grad.py and test_basic_modules.py) use random_post_comp=True with fully connect (so that the simulation results are the same as before). I could add tests for random_post_comp=True to test_connection.py -- would this then be enough coverage?

tests/test_connection.py
kyralianaka: This addresses bug #486 where connectivity_matrix_connect() would not work with e.g.
```
pre = net.cell(list(range(2, 5)))
post = net.cell(list(range(5, 7)))
```
but did when the pre range started at 0, e.g. 
```
pre = net.cell(list(range(4, 4)))
post = net.cell(list(range(4, 8)))
```

michaeldeistler:
@kyralianaka please request a review when ready.

jaxley/connect.py
```diff
@@ -44,33 +44,52 @@ def fully_connect(
...
+    # Get the indices of the connections, like it's a fully connected connectivity matrix
+    from_idx = np.repeat(range(0, num_pre), num_post)
+    to_idx = np.tile(
+        range(0, num_post), num_pre
+    )  # used only if random_post_comp is False
+
+    # Pre-synapse at the zero-eth branch and zero-eth compartment
+    global_pre_comp_indices = (
+        pre_cell_view.scope("local").branch(0).comp(0).nodes.index.to_numpy()
+    )  # setting scope ensure that this works indep of current scope
```
jnsbck, line 73: make sure to reset this after, in case the user is in global scope

jaxley/connect.py
```diff
@@ -80,45 +99,50 @@ def sparse_connect(
...
-    )
-    post_rows = post_cell_view.base.nodes.loc[global_post_indices]
-
-    # Pre-synapse is at the zero-eth branch and zero-eth compartment.
-    global_pre_indices = pre_cell_view.base._cumsum_nseg_per_cell[pre_syn_neurons]
-    pre_rows = pre_cell_view.base.nodes.loc[global_pre_indices]
+    num_pre = len(pre_cell_view._cells_in_view)
+    num_post = len(post_cell_view._cells_in_view)
+
+    # Generate random cxns without duplicates --> respects p but memory intesive if extremely large n cells
```
jnsbck, line 120: When I implemented the connect API, @michaeldeistler and me had discussed not wanting to do this, since this can be really bad for large, sparse networks. Did we change our mind about doing it this way?

jnsbck: I agree its cleaner and easier to parse, but not efficient.

jaxley/connect.py
```diff
@@ -44,33 +44,52 @@ def fully_connect(
...
+        pre_cell_view.scope("local").branch(0).comp(0).nodes.index.to_numpy()
+    )  # setting scope ensure that this works indep of current scope
+    pre_rows = pre_cell_view.select(nodes=global_pre_comp_indices[from_idx]).nodes
+
+    if random_post_comp:
+        # Randomly sample the post-synaptic compartments
+        global_post_comp_indices = (
+            post_cell_view.nodes.groupby("global_cell_index")
+            .sample(num_pre, replace=True)
+            .index.to_numpy()
```
jnsbck, line 81: This can potentially be split out no, since its used at least twice.

jaxley/connect.py
```diff
@@ -44,33 +44,52 @@ def fully_connect(
     pre_cell_view: "View",
```
jnsbck, line 44: should we rename `fully_connect` to `fully_connect_cells`? Might be less confusing, since otherwise connections are made comp 2 comp

jaxley/connect.py
```diff
@@ -44,33 +44,52 @@ def fully_connect(
...
-    global_post_indices = global_post_indices.reshape((-1, num_pre), order="F").ravel()
-    post_rows = post_cell_view.nodes.loc[global_post_indices]
-
-    # Pre-synapse is at the zero-eth branch and zero-eth compartment.
-    pre_rows = pre_cell_view.scope("local").branch(0).comp(0).nodes.copy()
-    # Repeat rows `num_post` times. See SO 50788508.
-    pre_rows = pre_rows.loc[pre_rows.index.repeat(num_post)].reset_index(drop=True)
+    # Get the indices of the connections, like it's a fully connected connectivity matrix
+    from_idx = np.repeat(range(0, num_pre), num_post)
+    to_idx = np.tile(
```
jnsbck, line 66: `if not random_post_comp:` would avoid overhead.

jaxley/connect.py
```diff
@@ -44,33 +44,52 @@ def fully_connect(
...
+        range(0, num_post), num_pre
+    )  # used only if random_post_comp is False
+
+    # Pre-synapse at the zero-eth branch and zero-eth compartment
+    global_pre_comp_indices = (
+        pre_cell_view.scope("local").branch(0).comp(0).nodes.index.to_numpy()
```
jnsbck, line 72: can also be done with `pre_cell_view.nodes.groupby("global_cell_index").first()["global_comp_index"]`, then you dont have to do all the scoping

jnsbck (changes requested):
Had a quick look and left some comments, not sure when you requested the review, if it was recent or months ago.

jaxley/connect.py
michaeldeistler: Good catch Jonas! Let's avoid building a `n x n` matrix.

kyralianaka:
Thanks for the feedback, I've done a bit of runtime testing and there is still some room for improvement, but my goal is still to make it as readable as possible (so that it's easy to write more functions like this) while not too slow. I'll work on this on Monday and think a bit more about restructuring

jaxley/connect.py
kyralianaka: Definitely was thinking about that, will make sure to include!

jaxley/connect.py
kyralianaka: I was looking into this for quite a while, and all of the ways that are memory efficient are extremely time intensive, and vice versa, so I came up with a way to split the cost a bit... generating blocks of random matrix at a time. I am still thinking of how it could be prettier, but the resource management seems to be good

kyralianaka:
An update on the timing: 
|             | PR    | original             |
|-------------|-------|----------------------|
| fc 0-0      | 2.485 s |                      |
| fc rand     | 2.205 s | 2.00 s                    |
| sparse 0-0  | 3.16 s  |                      |
| sparse rand | 3.24 s  | 10963.41 s (3 h) |

The random selection does actually seem to be faster now, so I'll just make that the default again unless that changes. Also shifting back towards the old fully connect which didn't have any problems other than not being so easy to parse.

jaxley/connect.py
kyralianaka: The function before created many duplicate connections between the same cells, and simply filtering these out would remove a lot of connections (inefficient) and make the p value entered somewhat meaningless (more problematic). Also p=1.0 did not create a fully connected graph, far from it I might add. Resource efficient methods for constructing sparse random graphs are highly sought after and typically trade off computation speed and memory, so I tried to write a function that balances the two. The function now isn't O(N^2) per se, the largest matrix it creates is 100x100. I can keep looking for better algorithms in the future, but for now the function at least fixes the bug of duplicate connections and restores the traditional meaning of p.

kyralianaka: I'm also thinking now that just removing sparse_connect() might not be so bad... I think we might have used connectivity matrix connect or something else in the RNN experiments anyway

jnsbck:
@kyralianaka what is happening here? :)

kyralianaka:
The state of this was that the sparse connect function was taking very very long to connect the cells, and it wasn't doing it a way one would expect in that it would often connect the same cells multiple times and p=1.0 would not fully connect the cells...

Then I did some digging and came up with a function for sparse connectivity that traded of memory (normally a full adj-mat would be needed) and time (assessing whether to connect cells one at a time). This was after a previous version of the function that generated expected connectivity patterns at the expense of generating the whole adj-mat. The last I heard though was that Michael wanted to revert back to the old function, maybe unless I could get append_multiple_synapses to work with sparse array types like csc or csr. The runtime problem of the old function could be solved by removing the random selection of a post synaptic compartment for each cell, so I made the random sampling its own function and an optional call of all of the fully_connect function as well. I think it's nice to have the option of knowing that the post-synaptic compartment is always zero (there was a use case for me a while ago) or randomly selecting it. This results in a slight increase in computational time for fully-connecting a large network (table above) which I might be able to reduce by shoving the random selection back into the function, but it doesn't save that much and I didn't think it'd be as clean. 

Anyway, what's left for me to do as I understand it is revert sparse-connect back to the old function without or with improved random sampling (because of the time problem) and write my concerns about the unexpected connections generated in the doc string. I'll try to push that here today, and you all can let me know if we should eventually merge this.

jnsbck:
sounds great, thanks for the update :) Even if we just make first-to-first default, it should probably make things faster and more predictable, which I like

kyralianaka:
New (hopefully final) times for connecting a network of 1000 cells with p=0.5 for the sparse connectivity:

|             | PR    | original             |
|-------------|-------|----------------------|
| fc 0-0      | 0.36 s |                      |
| fc rand     | 0.39 s | 0.39 s                    |
| sparse 0-0  | 0.20 s  |                      |
| sparse rand | 0.40 s  | 10963.41 s (3 h) |

the sparse random connectivity is faster now because I switched to using the pandas sample method instead of list comprehension, but 0-0 connectivity is default.

tests/jaxley_identical/test_grad.py
kyralianaka: Update: all those function tests now have both random_post_comp=True & False

michaeldeistler (approved):

Issue #485, opened 2024-11-07
Clamping of Synapse: Following Synapse might also be clamped


deezer257:
If I create a network where cells are connected via RibbonSynapses, and I apply a data clamp to one of the RibbonSynapses, not only is the current synapse clamped to the defined value, but also the next synapse in sequence (with an incremented index). An example is attatched.

Jax version: 0.4.35

[Synapses_example.zip](https://github.com/user-attachments/files/17657107/Synapses_example.zip)

jnsbck:
Hey, thanks for reporting this. Two things that would be really helpful to clarify this:
1. Did you check if this also applies to other synapses, i.e. `IonotropicSynapse`? If this is not the case I suspect this might be an issue in `jaxley_mech`.
2. If 1 is not the case, could you come up with a more minimal example, i.e. rm everything that's not strictly necessary but still produces this behavior. That would be really helpful to debug this.

deezer257:
I tried to clamp the parameters of the IonotropicSynapse, but the library doesn't accept that in the data_clamp:

```python
import matplotlib.pyplot as plt
import jaxley as jx
import jax
from jaxley.synapses import IonotropicSynapse
from jaxley.connect import connect
from jaxley_mech.channels.hodgkin52 import Leak, Na, K
import jax.numpy as jnp
import numpy as np

# Create a new dummy network
comp_opt = jx.Compartment()
branch_opt = jx.Branch(comp_opt, nseg=1)
cell_opt = jx.Cell([branch_opt], [-1])
net_opt = jx.Network(cells=[cell_opt] * 4)

# Insert the leak channel into the cell
net_opt.insert(Leak())
net_opt.insert(Na())
net_opt.insert(K())

# Connect the cells with the synapses
connect(net_opt.cell(0),  net_opt.cell(1), IonotropicSynapse())
connect(net_opt.cell(1), net_opt.cell(2), IonotropicSynapse())
connect(net_opt.cell(2), net_opt.cell(3), IonotropicSynapse())

# Set the conductance of the synapse to zero
net_opt.cell('all').set('IonotropicSynapse_gS', 0)

inputs = jnp.ones(1000)

net_opt.delete_recordings()
net_opt.delete_stimuli()

# Record the conductance of the synapses
net_opt.cell([0,1]).record("IonotropicSynapse_gS")
net_opt.cell([1,2]).record("IonotropicSynapse_gS")
net_opt.cell([2,3]).record("IonotropicSynapse_gS")

[...]

Issue #363, opened 2024-05-22
record synaptic currents


coschroeder:
at the moment it is not possible to record that

ntolley:
Is there any interest in adding functionality for recording extracellular fields (since they're calculated directly from these currents)?

I'm planning to start playing around with jaxley to simulate LFP/EEG, happy to try and contribute if there's interest in a dedicated API

michaeldeistler:
Yes, it would definitely be great to have this! Let me know if you need any input or would like to discuss, I would be happy to have a quick chat over zoom!

Michael

Aside: my guess is that achieving this will require to have access to x, y, z of postsynaptic compartments. Once #464 is merged, those can be accessed by doing:
```python
net._update_nodes_with_xyz()
net.copy_node_property_to_edges("x", "post")
net.edges
```

ntolley:
I'm actually realizing that the index on this dataframe does seem to correspond to `global_comp_index`! At least the index values repeat when adding recordings for multiple current types...

jnsbck:
Indeed, the index of whatever you are recording is written to `rec_index`, see https://github.com/jaxleyverse/jaxley/blob/32068e9e7d463f8170e1e86f02c46c18cbaaefeb/jaxley/modules/base.py#L1479-L1481. This can be the edge or comp index depending on whether the state is part of a synapse or channel.

PR #492, opened 2024-11-08
Fix synapse clamping

jnsbck:
Clamping of synaptic states was not tested and slipped through the cracks when #447 was merged.  This lead to several issues #485.

This PR fixes this and adds test coverage.


jaxley/integrate.py
```diff
@@ -118,11 +118,11 @@ def add_stimuli(
             if data_stimuli is not None:
                 externals["i"] = jnp.concatenate([externals["i"], data_stimuli[1]])
                 external_inds["i"] = jnp.concatenate(
-                    [external_inds["i"], data_stimuli[2].global_comp_index.to_numpy()]
+                    [external_inds["i"], data_stimuli[2].index.to_numpy()]
```
jnsbck, line 121: changed to `.index`, since `edges` does not have `global_comp_index`

jnsbck: ...and clamping of synapses needs the dataframe index.

jnsbck:
fyi @michaeldeistler 
I will merge this since all tests are passing and @kyralianaka  and @deezer257 rely on it. Please have a look nonetheless and poke me if I should touch some things up afterwards still.

Also @huangziwei I noticed we have #363 still open? I think this should be easy to do now that #487 is merged.

jaxley/modules/base.py
```diff
@@ -1638,23 +1660,28 @@ def delete_stimuli(self):
...
-            base_exts = self.base.externals
-            base_exts_inds = self.base.external_inds
-            if np.all(~keep_inds):
-                base_exts.pop(state_name, None)
-                base_exts_inds.pop(state_name, None)
+        all_externals = list(self.externals.keys())
+        if "i" in all_externals:
+            all_externals.remove("i")
+        state_names = all_externals if state_name is None else [state_name]
+        for state_name in state_names:
```
jnsbck, line 1669: Took the liberty of making state_name optional. If none is supplied `delete_clamps()` will just remove all clamps.

PR #493, opened 2024-11-08
Add View tutorial

jnsbck:
I added a tutorial to introduce the concepts behind Views and Modules, since I think understanding the basics is really helpful to work with Jaxley effectively and at the moment you'd have to figure it out yourself from the API docs.

This is a first draft. Your feedback would be much appreciated.

@kyralianaka @michaeldeistler.

### DIFF
--- a/docs/tutorials/01_morph_neurons.ipynb
+++ b/docs/tutorials/01_morph_neurons.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "9ad6d0c2",
+   "id": "e32c96c4",
    "metadata": {},
    "source": [
     "# Basics of Jaxley"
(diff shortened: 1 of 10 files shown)


michaeldeistler:
I will have a look later. Is it ok if I just push to the branch?

jnsbck:
Yes, fine with me.

michaeldeistler:
Awesome!

I think it's great to have a central place where all of these concepts are explained, and I think that users will find this very useful.

However, I would not suggest to make it the first tutorial---it is quite detailed and it is not very satisfying to end the first tutorial with "hey here is your network, but you cannot do anything with it yet".

So, I suggest the following (which I will push in a second, but we can ofc revert if you disagree):
- I copied some parts of your tutorial to `01_morph_neurons.ipynb`, in particular using `cell.nodes` and `cell.branch(0).nodes` to inspect the module.
- I added links to your new tutorial in the `01_` tutorial` and once in the `03_` tutorial
- I added your new tutorial as the first tutorial of the `Advanced tutorials` section (I created this new section in #464)

Let me know what you think!

PR #494, opened 2024-11-10
Add support for boolean indexing of cells/branches/comps

jnsbck:
This PR adds boolean indexing to Views. This is for convenience. Can be used to do stuff like:
```python
r_greater_1 = net.nodes.groupby("global_cell_index)["r"].mean() > 1
net[r_greater_1].nodes.vis()
# or
net.select(nodes=net.nodes["r"] > 1) # vs. net.select(nodes=net.nodes.index[net.nodes["r"] > 1])
```

### DIFF
--- a/tests/test_viewing.py
+++ b/tests/test_viewing.py
@@ -317,7 +317,7 @@ def test_view_supported_index_types(module):
         [0, 1, 2],
         np.array([0, 1, 2]),
         pd.Index([0, 1, 2]),
-        np.array([True, False, True]),
+        np.array([True, False, True] * 100)[: len(module.nodes)],
     ]
 
     # comp.comp is not allowed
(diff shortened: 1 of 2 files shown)


tests/test_viewing.py
```diff
@@ -329,9 +330,9 @@ def test_view_supported_index_types(module):
...
 
-            # for loc test float and list of floats
-            assert module.loc(0.0), "Failed for float"
-            assert module.loc([0.0, 0.5, 1.0]), "Failed for List[float]"
+        # for loc test float and list of floats
+        assert module.loc(0.0), "Failed for float"
+        assert module.loc([0.0, 0.5, 1.0]), "Failed for List[float]"
     else:
         with pytest.raises(AssertionError):
             module.comp(0)
```
michaeldeistler, line 338: Could we have three more tests:

1) for correctness of the boolean indexing.
2) For allowing boolean indexing of synapses.
3) Combining select with boolean indexing.

michaeldeistler (approved):
I love this! Let's have a few more tests, but good to go then!

PR #499, opened 2024-11-14
Refactor tests with fixtures

jnsbck:
This PR adds test fixtures for the modules that have to be re initialized again and again during testing. This should reduce overhead and make tests more readable. The fixtures are such that if i.e. `Branch(nseg=4)` was already used by a prev test, the fixture returns a copy of it, if not it is build from scratch and cached. Higher order modules, i.e. `Cell([Branch(nseg=4)], [-1,0,0])` that are build later can also make use of the pre-build submodules via the fixtures.  

This compliments #479 and contributes towards #449.


tests/conftest.py
```diff
@@ -0,0 +1,201 @@
...
+    def get_or_compute_swc2jaxley_params(
+        fname: str = None,
+        max_branch_len: float = 2_000.0,
+        sort: bool = True,
+        force_init: bool = False,
+    ):
+        dirname = os.path.dirname(__file__)
+        default_fname = os.path.join(dirname, "swc_files", "morph.swc")
+        fname = default_fname if fname is None else fname
+        if key := (fname, max_branch_len, sort) not in params or force_init:
```
michaeldeistler, line 196: this is the first time I have seen somebody actually use the walrus operator :D

tests/test_fixtures.py
```diff
@@ -0,0 +1,73 @@
...
+import warnings
+
+import pytest
+
+import jaxley as jx
+
+pytest.skip(allow_module_level=True)
+
+
+def test_module_retrieval(SimpleNet):
```
michaeldeistler, line 14: nice!

michaeldeistler (approved):
Amazing!

PR #500, opened 2024-11-14
Rework all tutorials

michaeldeistler:
- I removed tutorial `03_setting_parameters` as it had very little content and I want to keep the number of tutorials small. I added its main content into tutorials `01_` and `02_`.
- There was a bug in the dataloader (it only ran one epoch)
- make sure that all tutorials run and behave as expected


docs/tutorials/07_gradient_descent.ipynb
```diff
@@ -784,47 +792,50 @@
...
+    "\n",
     "for epoch in range(10):\n",
     "    epoch_loss = 0.0\n",
+    "    \n",
+    "    # Our simple dummy dataloader must be re-initialized at every epoch.\n",
+    "    dataloader = Dataset(inputs, labels)\n",
+    "    dataloader = dataloader.shuffle(seed=epoch).batch(batch_size)\n",
+    "\n",
     "    for batch_ind, batch in enumerate(dataloader):\n",
-    "        current_batch = batch[0].numpy()\n",
```
jnsbck, line 826: Could also be changed to  https://github.com/jaxleyverse/jaxley/blob/0df50b6aa7338ffe16943afbac410f2616b2e4bb/docs/tutorials/07_gradient_descent.ipynb#L826

docs/tutorials/07_gradient_descent.ipynb
```diff
@@ -784,47 +792,50 @@
...
    ],
    "source": [
+    "batch_size = 4\n",
+    "\n",
     "for epoch in range(10):\n",
     "    epoch_loss = 0.0\n",
+    "    \n",
+    "    # Our simple dummy dataloader must be re-initialized at every epoch.\n",
+    "    dataloader = Dataset(inputs, labels)\n",
+    "    dataloader = dataloader.shuffle(seed=epoch).batch(batch_size)\n",
```
jnsbck, line 834: Can we change this to
```
dataset = Dataset(inputs, labels)
for epoch in range(10):
    epoch_loss = 0.0
    
    # Our simple dummy dataloader must be shuffled at every epoch.
    dataloader = dataset.shuffle(seed=epoch).batch(batch_size)
```

jnsbck: alternatively, you can also increment the seed after the last yield or sth. than you would not have to re-init

jnsbck (approved):
Left two minor comments, but it looks good to me. Feel free to merge.

PR #509, opened 2024-11-15
nseg->ncomp for swc reader; set assign_groups=True

michaeldeistler:
This PR suggests a few things:

1) Rename `nseg` to `ncomp` when branches are initialized: `branch = jx.Branch(comp, ncomp=4)`
2) Rename `nseg` to `ncomp` in `jx.read_swc(..., ncomp=4)`
3) Rename `nseg` to `ncomp` throughout the entire codebase internally
4) Set `max_branch_len=None` by default. We now have `.set_ncomp(8)` to deal with long branches.
5) Set `assign_groups=True` by default.

All of these things are also reflected in the SWC tutorial. I also advertise `.set_ncomp` in the SWC tutorial now.


jnsbck:
I think we might want to move away from `nseg` altogether anyway, since if will be less confusing for the user. i.e. also change `cumsum_nseg`, `jx.Branch(ncomps=2)` etc. accordingly

jnsbck (approved):
My only high level comment is that we might wanna do this for all nseg. Otherwise good to go from my side.

Also, if we start working with deprecations now, we might wanna think about integrating this into the CI. See #516 for an idea of how to do this.

PR #514, opened 2024-11-19
Enable param sharing with differently sized groups

### LINKED ISSUE #501: Parameter sharing with different numbers of synapses
The new tutorial on advanced parameter sharing suggests that something like the following would be possible:
```
comp = jx.Compartment()
branch = jx.Branch(comp, nseg=1)
cell = jx.Cell(branch, parents=[-1])
net = jx.Network([cell for _ in range(3)])
connect(net.cell(0), net.cell(1), IonotropicSynapse())
connect(net.cell(0) ,net.cell(2), IonotropicSynapse())
connect(net.cell(1), net.cell(2), IonotropicSynapse())

net.copy_node_property_to_edges("global_cell_index")

df = net.edges
df["controlled_by_param"] = df["pre_global_cell_index"]
net.make_trainable("IonotropicSynapse_gS")
```
but this will throw an error because two synapses are controlled by one pre-synaptic cell and only one synapse by another pre-synaptic cell. 

This is a major limitation that prevents cells having the same (pre-)synaptic parameters but different numbers of post-synaptic partners. Maybe I am missing something as to how this parameter sharing should be done -- please let me know if so. In any case I will keep looking into it.

jnsbck:
This PR attempts to enable param sharing for controll groups with different numbers of members. This can be important for sharing params between branches with different numbers of segments or edges with different numbers of pre synaptic cells (see #501).

This also fixes #501.

todos:
- [x] add tests
- [x] rm asserts that prev. prevented the user from doing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ net.arrange_in_layers([3,3])
 net.vis()
 ```
 
+- Allow parameter sharing for groups of different sizes, i.e. due to inhomogenous numbers of compartments or for synapses with the same (pre-)synaptic parameters but different numbers of post-synaptic partners. (#514, @jnsbck)
+
 # 0.5.0
 
 ### API changes
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1075,14 +1075,24 @@ class Module(ABC):
         grouped_view = data.groupby("controlled_by_param")
         # Because of this `x.index.values` we cannot support `make_trainable()` on
         # the module level for synapse parameters (but only for `SynapseView`).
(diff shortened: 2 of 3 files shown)


jnsbck:
Surprisingly little had to be changed for this to work. I have added a test that checks make_trainable and set behave the same. Lemme know if any other tests are required.

michaeldeistler (approved):
Thank you, this is great!

jnsbck:
Thanks for adding the clarifying comment :)

jnsbck:
fyi @kyralianaka this has been merged now, please let me know if anything does not work as expected!

PR #518, opened 2024-11-20
rename global_pre_comps

jnsbck:
@michaeldeistler you can have a look at it, but I guess there is not much to see. If tests pass, I will just merge it. Hope that is fine :)


michaeldeistler (approved):
Thanks!

jnsbck:
missed a few replaces -.-, now it should pass

PR #520, opened 2024-11-20
rename `_update_nodes_with_xyz`

jnsbck:
renamed `_update_nodes_with_xyz` to `compute_compartment_centers` which also exposes it.
Also changed the toggle `update_nodes` in the movement methods to `False`, to not confuse the user if they move a module and suddenly the comp centers appear in nodes.


michaeldeistler (approved):
Perfect, thank you!

PR #521, opened 2024-11-20
add uninsert method

jnsbck:
This PR adds an `uninsert` method. 

I find the name a bit wonky tbh. Was also wondering if we should add a `disconnect` method to `connect.py`.

todos:
- [x] add tests

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1735,6 +1735,26 @@ class Module(ABC):
         for key in channel.channel_states:
             self.base.nodes.loc[self._nodes_in_view, key] = channel.channel_states[key]
 
+    def uninsert(self, channel: Channel):
+        """Remove a channel from the module.
+
+        Args:
+            channel: The channel to remove."""
+        name = channel._name
+        channel_names = [c._name for c in self.channels]
+        if name in channel_names:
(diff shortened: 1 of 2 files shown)


michaeldeistler:
Regarding the name: I am not a huge fan of `uninsert` either, but not sure about other options. Maybe we could go with `delete_channel()` (similar to `delete_recordings()`)?

jnsbck:
I like delete_channel much better! @michaeldeistler if you could have a quick look, that'd be great :)

jaxley/modules/base.py
```diff
@@ -1735,6 +1735,28 @@ def insert(self, channel: Channel):
...
+        name = channel._name
+        channel_names = [c._name for c in self.channels]
+        all_channel_names = [c._name for c in self.base.channels]
+        if name in channel_names:
+            channel_cols = list(channel.channel_params.keys())
+            channel_cols += list(channel.channel_states.keys())
+            self.base.nodes.loc[self._nodes_in_view, channel_cols] = float("nan")
+            self.base.nodes.loc[self._nodes_in_view, name] = False
+
+            # only delete cols if no other comps in the module have the same channel
```
michaeldeistler, line 1752: nice!

michaeldeistler (approved):
Amazing, thanks!

Issue #515, opened 2024-11-19
API for recording extracellular fields


ntolley:
As mentioned in #363, I'd be interested in adding the ability to record local field potentials (and possibly EEG/current dipole). Something important to decide on is what the API would look like.

At the most minimal level there wouldn't even be an API, but instead just a tutorial notebook showing how it can be calculated (perhaps with helper functions built into`utils`)

For an example of a real API, we use the following for[ hnn-core](https://github.com/jonescompneurolab/hnn-core):

```py
depths = list(range(-325, 2150, 100))
electrode_pos = [(135, 135, dep) for dep in depths]
net.add_electrode_array('shank1', electrode_pos)
```
For this repo I think `net.record_lfp()` might be a better fit considering the existing function names

Alternatively this could be merged with the `net.record()` functionality that currently exists. Something like 
```py
net.record(state='lfp', electrode_pos=electrode_pos)
```

This is a little confusing since the LFP isn't a state, and it'd add an obligatory argument to this method when `state='lfp'`. 

In any case would love to hear your thoughts on this!

jnsbck:
Cool that you are thinking about a `record_lfp` implementation! I think the ability to obtain lfps will be really useful. However, rather than adding this into the jaxley modules directly, I would be for making this functionality more of a util, i.e. `utils/lfp.py`, since I'd argue that lfps are not a core part of jaxley. In the long-run this could even become sth like `jaxley-lfp` if it ends up being super big. Also since `lfps` are essentially just a function of voltage (correct me if I am wrong), I think this should be easily separable from the rest of jaxley, i.e. via sth. like a `lfp_of_voltage(voltage, rec_pos, electrode_pos)`. Lemme know if you disagree @michaeldeistler @ntolley or have other .

jnsbck:
Cool, the membrane currents we can just record, for the synapses is currently not implemented, which why #363 is still open.

Regarding your Qs:
 1. Good catch! "i" is just the external current and is part of `_get_state_names` to be able reuse it in `stimulate` and `clamp`. In `record` it throws an error. We should either raise a more specific error in `record` or rm it from `_get_state_names` and append it manually in `clamp` / `stimulate`.
[...]

PR #523, opened 2024-11-20
[MRG] Add ability to record synaptic currents

### LINKED ISSUE #363: record synaptic currents

ntolley:
Closes #363 and #533

As discussed with @jnsbck in #515, this PR adds `synapse_current_names` as an attribute to the base module, and then appends this to the synapse states in `_get_state_names()`. The end goal is for currents to be recordable via:

```py
net.record('IonotropicSynapse_current')
```

Some lingering thoughts:
1) Is it worth changing "current" to "i" for consistency with the channel naming conventions?
2) It took me awhile to figure out that I needed to modify the `View` class. Will it be necessary to add a `self._synapses_in_view()` property for this to work properly?
3) Add some tests that check basic assumptions of the current recording?

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1210,7 +1210,10 @@ class Module(ABC):
         channel_states = [name for c in self.channels for name in c.channel_states]
         synapse_states = [name for s in self.synapses for name in s.synapse_states]
         membrane_states = ["v", "i"] + self.membrane_current_names
-        return channel_states + membrane_states, synapse_states + self.synapse_current_names
+        return (
+            channel_states + membrane_states,
+            synapse_states + self.synapse_current_names,
+        )
 
     def get_parameters(self) -> List[Dict[str, jnp.ndarray]]:
         """Get all trainable parameters.
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -2495,7 +2495,7 @@ class View(Module):
             ].unique()
             pre = base_edges["pre_global_comp_index"].isin(incl_comps).to_numpy()
             post = base_edges["post_global_comp_index"].isin(incl_comps).to_numpy()
-            possible_edges_in_view = base_edges.index.to_numpy()[(pre & post).flatten()]
+            possible_edges_in_view = base_edges.index.to_numpy()[(pre | post).flatten()]
             self._edges_in_view = np.intersect1d(
                 possible_edges_in_view, self._edges_in_view
             )
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -148,6 +148,7 @@ class Module(ABC):
         self.synapse_param_names = []
         self.synapse_state_names = []
         self.synapse_names = []
+        self.synapse_current_names: List[str] = []
 
         # List of types of all `jx.Channel`s.
         self.channels: List[Channel] = []
@@ -1209,7 +1210,7 @@ class Module(ABC):
         channel_states = [name for c in self.channels for name in c.channel_states]
         synapse_states = [name for s in self.synapses for name in s.synapse_states]
         membrane_states = ["v", "i"] + self.membrane_current_names
-        return channel_states + membrane_states, synapse_states
+        return channel_states + membrane_states, synapse_states + self.synapse_current_names
 
     def get_parameters(self) -> List[Dict[str, jnp.ndarray]]:
         """Get all trainable parameters.
@@ -2445,6 +2446,7 @@ class View(Module):
 
         self.channels = self._channels_in_view(pointer)
         self.membrane_current_names = [c._name for c in self.channels]
+        self.synapse_current_names = pointer.synapse_current_names
         self._set_trainables_in_view()  # run after synapses and channels
         self.num_trainable_params = (
             np.sum([len(inds) for inds in self.indices_set_by_trainables])
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -2495,8 +2495,9 @@ class View(Module):
             incl_comps = pointer.nodes.loc[
                 self._nodes_in_view, "global_comp_index"
             ].unique()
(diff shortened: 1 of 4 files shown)


ntolley:
FYI here's the currents and voltages from the [tutorial network](https://jaxley.readthedocs.io/en/latest/tutorials/02_small_network.html)

image

Seems reasonable to me? I'll admit I'm just guessing on the units for the synaptic current :smile:

jnsbck:
Nice! Just had a superficial look and it looks like not a lot of changes had to be made for the above to work. Units should be the standard current units if I am not mistaken (see [here](https://www.neuron.yale.edu/neuron/static/docs/units/unitchart.html)) so nA seems right. Regarding your lingerning thoughts:
1. I'd also be in favor of changing it to "i"
2. The `View` v `Module` has a small learning curve indeeed! You might not have to add a `_synapses_in_view`, but we should make sure the membrane currents are filtered by the synapses in view. sth like `[i for i, syn in zip(self.base.synapse_current_names, self.synapses) if syn._name in i]` should do the trick. Also for more intuition on View v Module see [jaxley api tutorial](https://github.com/jaxleyverse/jaxley/blob/main/docs/tutorials/00_jaxley_api.ipynb), (I assume you already read the docstrings :) )
3. Yes, we should definitely make sure this is tested well. (which I just noticed we don't do for channel_currents at the moment), maybe you can add a quick test for `i_HH` along with the synapse current to https://github.com/jaxleyverse/jaxley/blob/5e3637c26e03b433b9a003c26186ecea6d4596eb/tests/test_record_and_stimulate.py#L57 other than testing that it works, I don't know what would be sensible to check though. @michaeldeistler any ideas?

Lemme me know when you want a review :)

michaeldeistler:
Awesome, thanks a lot for getting this started @ntolley!

I think the following would be important to test:

1) It plays nicely with [advanced indexing of synapses](https://jaxley.readthedocs.io/en/latest/tutorials/09_advanced_indexing.html). E.g., to record only synaptic currents of for pre-synaptic neuron 3 and 4:
```python
fully_connect(net.cell("all"), net.cell("all"), IonotropicSynapse())
net.copy_node_property_to_edges("global_cell_index")
df = net.edges
df = df.query("pre_global_cell_index in [3, 4]")
net.select(edges=df.index).record("Ionotropic_s")
```

2) If there are multiple kinds of synapses (e.g. `IonotropicSynapse` and `TanhSynapse`), then there is no mixup between them. E.g.:
```python
fully_connect(net.cell("all"), net.cell("all"), IonotropicSynapse())
fully_connect(net.cell("all"), net.cell("all"), TanhSynapse())
df = net.edges
df = df.query("SynapseType == 'IonotropicSynapse'")
net.select(edges=df.index).record("Ionotropic_s")
```

Let me know if anything is unclear!

jaxley/modules/base.py
```diff
@@ -2422,7 +2426,8 @@ def __init__(
         )
 
         self.channels = self._channels_in_view(pointer)
-        self.membrane_current_names = [c._name for c in self.channels]
+        self.membrane_current_names = [c.current_name for c in self.channels]
```
ntolley, line 2429: Think this was a bug, it wasn't actually possible to record channels currents from a `View` because of the naming

ntolley:
Got sent down a rabbit hole that led to #533, unfortunately edges aren't being updated in the `View` so that's going to need to be fixed to actually have the correct synpases be filtered out.

jaxley/modules/base.py
```diff
@@ -2490,7 +2495,7 @@ def _set_inds_in_view(
             ].unique()
             pre = base_edges["pre_global_comp_index"].isin(incl_comps).to_numpy()
             post = base_edges["post_global_comp_index"].isin(incl_comps).to_numpy()
-            possible_edges_in_view = base_edges.index.to_numpy()[(pre & post).flatten()]
+            possible_edges_in_view = base_edges.index.to_numpy()[(pre | post).flatten()]
```
ntolley, line 2498: @michaeldeistler @jnsbck this is the culprit for #533 ! Essentially if you're indexing `net.cell(0)`, then it's impossible for a view of that cell to include the compartments of another cell (unless it's an autapse)

I think this is a question of how you want this to behave. I've changed it to an "or" condition, so `net.cell(0).edges` will include both synapses onto this cell, as well as synapses it makes with other cells

For the purposes of this PR, it would make sense to only include the edges of the post synaptic cell, however you could argue for different expected behaviors.

jaxley/modules/base.py
```diff
@@ -1207,9 +1208,14 @@ def _get_state_names(self) -> Tuple[List, List]:
 
         Returns states seperated by comps and edges."""
         channel_states = [name for c in self.channels for name in c.channel_states]
-        synapse_states = [name for s in self.synapses for name in s.synapse_states]
+        synapse_states = [
+            name for s in self.synapses if s is not None for name in s.synapse_states
```
ntolley, line 1212: the `not None` filter is necessary since `self.synapses` is populated with `[None]` under certain conditions

I'll admit I don't totally understand why it's necessary just yet, although I saw the comment about it preserving indexing :smile:

jaxley/modules/base.py
ntolley: I decided to just go forward with only appending the post synaptic target so that I could finish the tasks for this PR. We can discuss if that's actually the desired behavior for the `View`

ntolley:
I think I've fixed most of the issues! The new tests are passing so this is ready for a review

michaeldeistler (approved):
This is awesome, thanks a ton!

A few comments below, mostly minor things. I suggest we stick to `pre & post` for now, we could change this behavior in another PR if we want to. Also, please add an entry in the changelog describing the new feature and the bug fixes :)

Good to go once changes are addressed! Thanks again!

michaeldeistler (approved):
LGTM, thanks a lot!

PR #537, opened 2024-12-02
Ensure Changelog is updated with every PR.

jnsbck:
I also like the suggestion of @ntolley to make updating the changelog part of PRs. To avoid having to remind contributors on every PR this could be enforced via github actions. I wrote a little check to do this.

(Quick test: changelog not updated -> tests should fail)


ntolley:
Definitely a great idea to automate the check! Are you thinking to leave it to the discretion of the maintainer if it's appropriate to ignore the check? (e.g. a typo fix)

jnsbck:
Currently this would be the case, but I am also not so happy about the current implementation at the moment. I also thought about adding a separate workflow i.e. "chores" and incl. isort, black, license header and changelog in there. This would be much cleaner and easier to quickly check for the maintainer as well.

jnsbck:
I did some refactoring:
- Tests/Chores checks for isort, license headers, changelog, black
- Tests/Pytest runs tests
- Tests/Regression Tests runs regression tests only if tests were succesful.

jnsbck:
@michaeldeistler thoughts?

michaeldeistler (approved):
Sorry, I had missed this one. I like the idea!

PR #538, opened 2024-12-02
Add first set of baselines

jnsbck:
This PR adds an initial set of baselines and verifies that baselines can be updated.


michaeldeistler (approved):
wuhuu, thanks a lot!

PR #539, opened 2024-12-03
Small refactor of plotting

### LINKED ISSUE #465: Inconsistent use of Dict and **kwargs for plotting functions [Bug]
In some places, the additional arguments that should be passed to matplotlibs plotting functions are passed via a dict, sometimes the function interface wants a dict.

For example, running:

```
import jaxley as jx

comp = jx.Compartment()
comp.compute_xyz()
comp.vis(type="comp", morph_plot_kwargs={"facecolor":'red'})
```

gives a Type Error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[...]/jaxley/modules/base.py", line 1914, in vis
    return plot_comps(self, dims=dims, ax=ax, col=col, **morph_plot_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: plot_comps() got an unexpected keyword argument 'facecolor'
```

Either it has to be changed there to `comp_plot_kwargs=morph_plot_kwargs` or the interface has to change to the **kwargs style.

jnsbck:
Fixes #465 and refactors some plotting functionality. 
Changes:
- rm `morph_plot_kwargs`, rely on `**kwargs` instead.
- renamed kwarg `"col"` -> `"color"` for consistency with pyplot.
- Refactor `Network.vis`
 - cleanup of code (use new View functionality) and removal of unnecessary kwargs
 - rm `_to_graph` since it is obsolete now (replaced by scatter, which might be not as nice as the graph plotting, can be added back in though.). Since networkx is not used currently, this could also be removed as a dependency (until #355 makes it in at least)
 - disentangle moving and visualizing of cells, by adding `arrange_in_layers`. I think this is more clear and seperates out the additional kwargs of `vis`.

To get the same as `net.vis(layers=[3,3])` one now has to do.
```python
net.arrange_in_layers([3,3])
net.vis()
```

This now also works for both `detail="full"` and `detail="point"` and even for `dims=[0,1,2]` (not that this would be necessary). In general `network.vis` should feel more similar to `cell.vis`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ net.record("i_IonotropicSynapse")
   - Regression tests can be done locally by running `NEW_BASELINE=1 pytest -m regression` i.e. on `main` and then `pytest -m regression` on `feature`, which will produce a test report (printed to the console and saved to .txt).
   - If a PR introduces new baseline tests or reduces runtimes, then a new baseline can be created by commenting "/update_regression_baselines" on the PR.
 
+- refactor plotting (#539, @jnsbck).
+  - rm networkx dependency
+  - add `Network.arrange_in_layers`
+  - disentangle moving of cells and plotting in `Network.vis`. To get the same as `net.vis(layers=[3,3])`, one now has to do:
+```python
+net.arrange_in_layers([3,3])
+net.vis()
+```
+
 # 0.5.0
 
 ### API changes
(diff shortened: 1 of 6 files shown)


jnsbck:
regression test will get an update see #540, which should make it less annoying to get them to pass.

michaeldeistler (approved):
Thanks for this, all looks good!

jnsbck:
fyi @kyralianaka in case you are using this.

PR #541, opened 2024-12-04
update tutorials to adapt new plotting API

michaeldeistler:
- change `col` to `color`
- update how `layers` are built

See also #542 on more clean up to be done for this.


jnsbck (approved):
looks good

Issue #543, opened 2024-12-04
Regression tests not working reliably


jnsbck:
Despite putting in a lot of work, the regression tests are not working reliably in the CI. I have done some reading and found a few hints to why this is. Mainly the github CI is not very reliably. See [here](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks/) and [ħere](https://pythonspeed.com/articles/consistent-benchmarking-in-ci/), one can expect large deviations (up to 50%) when running comparing the same code against itself in the CI. ![image](https://github.com/user-attachments/assets/816b9610-e75c-4375-98b5-a7122ca3f624). 

This means we might want to check for regressions differently:
1. only do it on demand (i.e. via github comment)
2. compute baseline at test time

I have also been looking into [asv](https://asv.readthedocs.io/en/latest/index.html), which looks really cool and would be something to implement during a hackathon maybe.

PR #546, opened 2024-12-04
Update Regression CI

jnsbck:
Since maintaining a baseline is not possible (see #543 for context), I have modified the regression CI to run baselines on main and tests on the PR branch in one go. The trigger for this is now a comment with "/test_regression".
This also makes skipping the regression tests automatic when there is no baseline_file around, rather than having to add `pytest -m "not regression"` every time.

Reuses a lot of code from before. Works for the moment. Long term we could look at using `asv` instead. (see #543)


jnsbck:
@michaeldeistler could you approve this? This should make the CI pass again.

michaeldeistler (approved):

PR #552, opened 2024-12-12
Explictly import Na,K,Leak from pospischil

michaeldeistler:
I got feedback that it is unintuitive that inserting `HH()` is different from inserting `Na(), K(), Leak()` separately. As a first step towards making this more clear, I am now importing these channels from `jx.channels.pospischil` explicitly.

### DIFF
--- a/docs/tutorials/00_jaxley_api.ipynb
+++ b/docs/tutorials/00_jaxley_api.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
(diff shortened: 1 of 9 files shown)


jnsbck:
I guess they are different from the HH() channels though. I think if might be easier to just split the HH channels apart, we can still keep HH() for convenience though where HH() just combines hh.Na(), hh.K(), hh.Leak().

michaeldeistler:
Replaced by #766

PR #561, opened 2024-12-20
WIP: Vectorize recording during `integrate`

ntolley:
While testing the recording from many _many_ states, I noticed I was experiencing some serious performance hits during simulation. @jnsbck suggested that this may be due to a for loop over recordings that occurs during the `integrate` call. This is an attempt to vectorize that update only indexing each unique state once.

### DIFF
--- a/jaxley/integrate.py
+++ b/jaxley/integrate.py
@@ -231,8 +231,12 @@ def integrate(
 
     if module.recordings.empty:
         raise ValueError("No recordings are set. Please set them.")
-    rec_inds = module.recordings.rec_index.to_numpy()
-    rec_states = module.recordings.state.to_numpy()
+    recording_df = module.recordings.reset_index(drop=True)
+    rec_states, rec_inds, sort_inds = list(), list(), list()
+    for state, df_group in recording_df.groupby("state"):
+        rec_states.append(state)
+        rec_inds.append(df_group.rec_index.to_numpy())
+        sort_inds.extend(df_group.index.to_list())
 
     # Shorten or pad stimulus depending on `t_max`.
     if t_max is not None:
@@ -260,7 +264,7 @@ def integrate(
 
     def _body_fun(state, externals):
         state = step_fn(state, all_params, externals, external_inds, delta_t)
-        recs = jnp.asarray(
+        recs = jnp.concatenate(
             [
                 state[rec_state][rec_ind]
                 for rec_state, rec_ind in zip(rec_states, rec_inds)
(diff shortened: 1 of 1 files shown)


jaxley/integrate.py
```diff
@@ -311,4 +315,5 @@ def _body_fun(state, externals):
         nested_lengths=checkpoint_lengths,
     )
     recs = jnp.concatenate([init_recording, recordings[:nsteps_to_return]], axis=0).T
+    # recs = recs[sort_inds, :]  # Sort recordings back to order that was set by user.
```
ntolley, line 318: This is a bit funky, when calling `df.groupby` you may be changing the order of the original `module.recordings`

This is meant to take care of that by putting things back to the original order, but uncommenting this line causes tests to fail so something in my logic is wrong

ntolley: nevermind! silly mistake, needed to use `argsort`

ntolley:
Here's some code to see how recordings impact speed: the comparison is pretty extreme but it gets the point across:
```py
import jaxley as jx
import time
from jaxley.synapses import IonotropicSynapse
from jaxley.connect import fully_connect
from jax import config
config.update("jax_enable_x64", True)
config.update("jax_platform_name", "cpu")

cell = jx.Cell()
cell.insert(Na())

sim_time_list, array_time_list = list(), list()

net = jx.Network([cell for _ in range(100)])
fully_connect(net, net, IonotropicSynapse())

params = net.get_parameters()

# Small number of recordings (4)
net.delete_recordings()
net.cell(range(2)).record('i_IonotropicSynapse')

start_time = time.time()
v = jx.integrate(net, params=params, t_max=10.0, delta_t=0.025)
simulate_time = time.time() - start_time
print(simulate_time)

# Huge number of recordings (10,000)
net.delete_recordings()
net.record('i_IonotropicSynapse')

start_time = time.time()
v = jx.integrate(net, params=params, t_max=10.0, delta_t=0.025)
simulate_time = time.time() - start_time
print(simulate_time)
```

ntolley:
And finally here's my timing results for the two different branches. There's still a slight slowdown with more recordings, but it's an order of magnitude faster when recording from 10,000 synapses so I'd say it's an improvement :smile: 

|  | `main` | `record_speedup` |
|--------|--------|--------|
| 4 recordings| 0.51 | 0.50 |
| 10,000 recordings | 10.9 | 1.15 |

ntolley:
@michaeldeistler @jnsbck since this is mainly a performance boost I'm not sure how it should be tested. I feel like testing the execution time directly could be very brittle for running tests locally. Unless you have some ideas, perhaps it isn't necessary?

ntolley:
Unfortunately that performance hit does scale with time, here's the results for a 100 ms simulation
|  | `main` | `record_speedup` |
|--------|--------|--------|
| 4 recordings| 0.55 | 0.63 |
| 10,000 recordings | 16.33 | 6.85 | 

So there's still some optimizations to be made...

PR #569, opened 2025-01-20
Windows dtype correction

Kartik-Sama:
I had to add a few more checks to ensure the np_dtype was being instantiated correctly, to avoid the assertion check. The changes are simple! Apologies for the messy commit history, I was cleaning it up but the branch already got merged so can't help it now.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -392,8 +392,7 @@ class Module(ABC):
             array of indices of shape (N,)"""
         if is_str_all(idx):  # also asserts that the only allowed str == "all"
             return idx
-
-        np_dtype = np.int64 if dtype is int else np.float64
+        np_dtype = np.dtype(int).type if dtype is int else np.dtype(float).type
         idx = np.array([], dtype=dtype) if idx is None else idx
         idx = np.array([idx]) if isinstance(idx, (dtype, np_dtype)) else idx
         idx = np.array(idx) if isinstance(idx, (list, range, pd.Index)) else idx
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -392,7 +392,11 @@ class Module(ABC):
             array of indices of shape (N,)"""
         if is_str_all(idx):  # also asserts that the only allowed str == "all"
             return idx
-        np_dtype = np.dtype(int).type if dtype is int else np.dtype(float).type
+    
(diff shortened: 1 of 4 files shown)


jaxley/modules/base.py
```diff
@@ -405,7 +409,12 @@ def _reformat_index(self, idx: Any, dtype: type = int) -> np.ndarray:
             dim = shape[np.where(which_idx)[0][0]]
             idx = np.arange(dim)[idx]
         assert isinstance(idx, np.ndarray), "Invalid type"
-        assert idx.dtype in [np_dtype, bool], "Invalid dtype"
+        assert idx.dtype in [np_dtype, bool], (
+            "Invalid dtype, found"
+            + str(idx.dtype)
+            + " instead of "
+            + str([np_dtype, bool])
+        )
```
jnsbck, line 417: one could fmt this using fstrings.

jaxley/modules/base.py
```diff
@@ -392,7 +392,11 @@ def _reformat_index(self, idx: Any, dtype: type = int) -> np.ndarray:
             array of indices of shape (N,)"""
         if is_str_all(idx):  # also asserts that the only allowed str == "all"
             return idx
-        np_dtype = np.dtype(int).type if dtype is int else np.dtype(float).type
+
+        if isinstance(idx, np.ndarray) and np.issubdtype(idx.dtype, np.number):
```
jnsbck, line 396: I just noticed, we could include a few more specific test cases for int/float np dtypes [here](https://github.com/jaxleyverse/jaxley/blob/8c7750a819469d7ef019e5385bde1b05133ee17b/tests/test_viewing.py#L292).

jnsbck (approved):
Thanks again. Good to go from my side once the comments I left are adressed :)

jaxley/modules/base.py
Kartik-Sama: I wasn't sure what test cases to add to check the float `np.float` dtypes. From the function description, it seems floats can be used for calling in the `loc` function, but I don't fully understand it.

Other than that I did add a few test cases to validate if everything is running with any integer dtypes.

michaeldeistler:
Hm, sorry to bother you again, but somehow the line you added to the `CHANGELOG.MD` got lost :D Could you add it back in?

PR #570, opened 2025-01-28
Update pyproject.toml

michaeldeistler:
This PR pins the jaxley versions to be older than `0.4.32`.

I realized that, with version jax==0.4.32 or newer, simulation time is 10x slower and gradient time is 5x slower as compared to older versions of JAX 🤯 

From the CHANGELOG of JAX:
```
This release of jaxlib switched to a new version of the CPU backend, which should compile faster
and leverage parallelism better. If you experience any problems due to this change, you can
temporarily enable the old CPU backend by setting the environment variable 
XLA_FLAGS=--xla_cpu_use_thunk_runtime=false. If you need to  do this, please file a JAX
bug with instructions to reproduce.
```
Indeed, os.environ['XLA_FLAGS'] = '--xla_cpu_use_thunk_runtime=false' at the beginning of the notebook fixes it also for newer versions of JAX.

See also [this issue that I created on the JAX repo](https://github.com/jax-ml/jax/issues/26145).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,11 @@ requires-python = ">=3.10"
 readme = "README.md"
 keywords = ["neuroscience", "biophysics", "simulator", "jax"]
 dependencies = [
-    "jax",
+    # See #571 for the pin on the JAX version. Jaxley runs with newer versions of JAX
+    # but is much slower on CPU. Newer versions can be made equally fast by setting
+    # `os.environ['XLA_FLAGS'] = '--xla_cpu_use_thunk_runtime=false'` at the beginning
+    # of a jupyter notebook.
+    "jax <= 0.4.31",
     "matplotlib",
     "numpy",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ requires-python = ">=3.10"
 readme = "README.md"
 keywords = ["neuroscience", "biophysics", "simulator", "jax"]
 dependencies = [
-    "jax",
+    "jax <= 0.4.31",
```
jnsbck, line 29: Let's link this issue in a comment here, in case ppl wonder / wanna use newer jax versions.

jnsbck (approved):
This is a bit annoying but I am also for fixing the version for now.

Have a look at the comment above, otherwise good to go.

jnsbck:
Can confirm also happening on my end

PR #574, opened 2025-01-29
WIP: Allow vectors to be used with `.data_set()`

### LINKED ISSUE #573: BUG: `.data_set()` function does not work with vectors
Fairly easy to reproduce, if you try to pass a vector for `data_set()` you get the following:
```py
import jaxley as jx
import numpy as np

cell = jx.Cell()
cell.insert(Na())
net = jx.Network([cell for _ in range(5)])
param_state = None
param_state = net.data_set('Na_gNa', np.repeat(1.0, 5), param_state)
net.record('v')
s = jx.integrate(net, t_max=10, param_state=param_state)
```
```
ValueError: Incompatible shapes for broadcasting: (5, 1) and requested shape (1, 5)
```

Seems like a simple enough change, will open a PR to try and fix this

ntolley:
Closes #573. Only change is swapping the dimensions when squeezing the `params` used for the `.set()`
 @michaeldeistler @jnsbck is there anything I'm missing? Hopefully this isn't breaking something somewhere else!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ net.vis()
 ### Bug fixes
 - Fixed inconsistency with *type* assertions arising due to `numpy` functions returning different `dtypes` on platforms like Windows (#567, @Kartik-Sama)
 
+- Fixed bug where `.data_set()` could not use vectors as input (#574, @ntolley)
+
 # 0.5.0
 
 ### API changes
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1316,9 +1316,9 @@ class Module(ABC):
             if key in params:  # Only parameters, not initial states.
                 # `inds` is of shape `(num_params, num_comps_per_param)`.
                 # `set_param` is of shape `(num_params,)`
-                # We need to unsqueeze `set_param` to make it `(num_params, 1)` for the
-                # `.set()` to work. This is done with `[:, None]`.
-                params[key] = params[key].at[inds].set(set_param[:, None])
+                # We need to unsqueeze `set_param` to make it `(1, num_params)` for the
+                # `.set()` to work. This is done with `.reshape(1, -1)`.
+                params[key] = params[key].at[inds].set(set_param.reshape(1, -1))
 
         # Compute conductance params and add them to the params dictionary.
         params["axial_conductances"] = self.base._compute_axial_conductances(
--- a/tests/test_make_trainable.py
+++ b/tests/test_make_trainable.py
@@ -447,15 +447,10 @@ def test_data_set_vs_make_trainable_network(SimpleNet):
 
 def test_data_set_vector(SimpleNet):
     net = SimpleNet(2, 2, 4)
-    net.set(np.repeat(1.0, 16))
-    with pytest.raises(ValueError, match="Incompatible shapes for broadcasting"):
-        net.cell(range(2)).set("radius", np.repeat(1, 16))
+    net.set("radius", np.repeat(1.0, 16))
 
     param_state = None
     param_state = net.data_set("length", np.repeat(1.0, 16), param_state)
-    param_state = net.cell(range(2)).data_set(
-        "radius", np.array([0.1, 0.2]), param_state
-    )
     net.record("v")
     jx.integrate(net, t_max=10, param_state=param_state)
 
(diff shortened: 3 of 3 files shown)


ntolley:
Also while working on this I realized the the regular `net.set()` function has some peculiar behavior as well. For instance the following code passes without any error when running the tests:
```py
net = SimpleNet(2, 2, 4)
net.set(np.repeat(1.0, 16))  # no error
net.cell(range(2)).set("radius", np.repeat(1.0, 16))  # no error
net.cell(range(2)).set("radius", np.repeat(1.0, 2))  # error
```
I've included tests for this which is why the PR is failing, they should pass if I limit the tests to just the `data_set()` functionality.

I may need some guidance on where to look to proceed. I imagine this is possible since `net.cell(range(2)).make_trainable(...)` has no issues with views

michaeldeistler:
Hi Nick! Thanks for the PR!

I think `set` behaves as I would have expected. Currently, `set` must either get a scalar value or an array with as many entries as there are compartments being set.

```python
net = SimpleNet(2, 2, 4)
```
creates a net of two neurons with 2 branches each, and each branch having 4 compartments. Then, `net.cell(range(2))` selects both cells => 16 compartments. This is why the first line passes.

I agree that it would be nice to support the second line. It's definitely possible, but it would require extra logic for the `set` method to understand that it should assign the two values to all compartments in the two cells.

That being said: since we currently do not support this yet, please remove the test.

I agree that there is something off with `data_set` (your example in the issue should be working).

ntolley:
Will do! Also I spoke to soon, this does indeed break a few things, keep at it...

michaeldeistler:
Yes, I also just found a small example that breaks after the change:
```python
import jaxley as jx
import numpy as np

comp = jx.Compartment()
branch = jx.Branch(comp, ncomp=4)
cell1 = jx.Cell(branch, parents=[-1])
cell2 = jx.Cell(branch, parents=[-1, 0, 0])
net = jx.Network([cell1, cell2])

net.cell("all").make_trainable("radius")
params = net.get_parameters()
net.record('v')
s = jx.integrate(net, t_max=10, params=params)
```

I think the best place to fix it would be [here](https://github.com/jaxleyverse/jaxley/blob/e6f7ab096533cebde115caf6a5e5537f353af45d/jaxley/modules/base.py#L874-L876), but it requires quite a bit of logic: If the number of compartments matches the number of values, then `inds` should be made into shape `(num_inds, 1)` [here](https://github.com/jaxleyverse/jaxley/blob/e6f7ab096533cebde115caf6a5e5537f353af45d/jaxley/modules/base.py#L874), not into shape `(1, num_inds)` (which happens with the `np.atleast2d`)

chaseking:
Any updates on this? The param state being added in [`data_set`](https://github.com/jaxleyverse/jaxley/blob/main/jaxley/modules/base.py#L899) doesn't seem to have the expected shape. The comment [here](https://github.com/jaxleyverse/jaxley/blob/main/jaxley/modules/base.py#L1347-L1348) indicates the `"indices"` entry is shape `(num_params, num_comps_per_param)` and the `"val"` entry is shape `(num_params,)`. The problem with the [original code](https://github.com/jaxleyverse/jaxley/blob/main/jaxley/modules/base.py#L899) is that atleast_2d adds singleton leading axes, rather than trailing. Something like this should fix it, and avoid the breaking changes outlined above?
```
"indices": jnp.asarray(viewed_inds[not_nan]).reshape(-1, 1),  # shape (n_comp, 1)
```

michaeldeistler:
Yes, I think this would work. I am not fully sure that it will also work with parameter sharing, but I think we have a decent set of tests for this that should indicate issues. Feel free to just try it out! (tagging @kyralianaka because she is also interested in this issue).

ntolley:
Thank you for the suggestion @chaseking! I'll definitely give it a try

For context I ran into this problem trying to implement an evolutionary algorithm for jaxley simulations, my workaround was to manually modify trainable values in the `params` dictionary

michaeldeistler:
Closing this in favor of #606. Thanks for the efforts @ntolley! Let us know if anything is not working after the fixes from #606.

Michael

ntolley:
Glad to see you guys were able to resolve it! Apologies for dropping off suddenly, I'm transitioning to thesis defense mode so my open source activity is ramping down for a bit. Hoping to resume once I've successfully defended (and after some vacation :smile:)

michaeldeistler:
No worries at all @ntolley! Best of luck with the defense!

PR #578, opened 2025-02-07
deprecate nseg and bump version

jnsbck:
Deprecation helper should have raised when bumping the version, but did not trigger. This fixes that, deprecates nseg and bumps the version to `0.6.1`.


michaeldeistler (approved):
Thanks! Please also change the version in `pyproject.toml`. Good to go then!

michaeldeistler:
And please add a line to the changelog.

PR #579, opened 2025-02-10
Make sphinx configuration file explicit

michaeldeistler:
The most recent readthedocs documentation (v0.6.0) did not build because we now require an explicit reference to the sphinx config file. See [here](https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/) for more info.

With this change, v0.6.1 should build successfully.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,9 @@
 # RTD configuration file version
 version: 2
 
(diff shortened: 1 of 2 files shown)


jnsbck (approved):
looks good, will rebase #578 when merged

PR #580, opened 2025-02-10
replace nseg in tutorials

jnsbck:
when making the release for 0.6.1, I unfortunately missed that the notebooks still contained mentions of `nseg`. This is fixed here now. Not sure if we should amend the last release...


michaeldeistler (approved):
Thanks for fixing this!

Let's make a 0.6.2 release once this is merged.

PR #584, opened 2025-02-13
Fix for rendering types of nodes in swc file

Kartik-Sama:
When a new branch is encountered, we need the previous branch's type and nodes in the resulting list of branches and branch types.

In the older version, the handling of the first branch made things complicated, where the type of branch was being updated at the start of the branch while the branch was being added at the end of the branch. Since this was creating many corner cases, the issue in #582 was being observed. 

For the following test case

> 
> # id,type,x,y,z,r,pid
> 1 1 0.0 0.0 0.0 0.1 -1
> 2 2 0.0 0.0 1.0 0.1 1
> 3 3 0.0 0.0 2.0 0.1 1
> 

The output of `cell.base.groups` with `n_comp = 2`,
- Previously : `{'soma': array([0, 1]), 'basal': array([2, 3, 4, 5])}`
- With this change : `{'soma': array([0, 1]), 'axon': array([2, 3]), 'basal': array([4, 5])}`

### DIFF
--- a/jaxley/__version__.py
+++ b/jaxley/__version__.py
@@ -1,6 +1,6 @@
 # This file is part of Jaxley, a differentiable neuroscience simulator. Jaxley is
 # licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
 
-VERSION = (0, 6, 1)
+VERSION = (0, 6, 2)
 
 __version__ = ".".join(map(str, VERSION))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "Jaxley"
-version = "0.6.1"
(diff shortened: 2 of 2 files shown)


jnsbck:
Thanks for looking into this and adding a fix. If this turns out to be too complicated, deprecating the custom reader might also be an option. What do you think @michaeldeistler?

michaeldeistler:
I wold prefer to still keep the old SWC reader, for two reasons:
- the old one has been stress-tested a bit more
- the old one is faster

jnsbck:
In that case we should definitely make sure it is fixed and potentially add a few more morphologies to the test suite (w.o. Simulating them, just importing and asserting that lengths, radii, and xyz cords match those of neuron)

Didn't realise it's much slower, thought they were about the same. I am sure it could be made much faster in principal.

PR #594, opened 2025-02-25
Bump version, prepare release 0.7.0

michaeldeistler:


jnsbck:
noticed that publishing to pypi failed. not sure but maybe we also need to set up a publisher on pypi?
image

PR #597, opened 2025-02-26
Fixes for `groups` after `.set_ncomp()`

### LINKED ISSUE #560: `set_ncomp` has to propagate to groups
When using `set_ncomp` the `groups` get messed up because they do not get updated.

michaeldeistler:
Closes #560 

In #560, we identified that groups do not get updated automatically when `set_ncomp()` is run. This PR fixes that by making `groups` a first-class citizen: Group identities are listed as a boolean column in `.nodes` now. With this, we need no further special treatment of `set_ncomp()`. In addition, we can get rid of special treatment of groups in the `graph` swc reader.


jaxley/modules/base.py
```diff
@@ -118,7 +118,7 @@ def __init__(self):
         self.total_nbranches: int = 0
         self.nbranches_per_cell: List[int] = None
 
-        self.groups = {}
+        self.group_names: List[str] = []
```
jnsbck, line 121: I see the reasons for keeping track of group names outside of `nodes`, but I slightly prefer the old `groups` style dictionary.

jaxley/modules/base.py
```diff
@@ -205,12 +205,9 @@ def __getattr__(self, key):
...
         # intercepts calls to groups
-        if key in self.base.groups:
-            view = (
-                self.select(self.groups[key])
-                if key in self.groups
-                else self.select(None)
-            )
+        if key in self.base.group_names:
+            inds = self.nodes.index[self.nodes[key]].to_numpy()
+            view = self.select(inds)
```
jnsbck, line 210: select can also take the boolean of self.nodes[key] directly :)

jaxley/modules/base.py
```diff
@@ -1242,12 +1251,12 @@ def add_to_group(self, group_name: str):
         Args:
             group_name: The name of the group.
         """
-        if group_name not in self.base.groups:
-            self.base.groups[group_name] = self._nodes_in_view
+        if group_name not in self.base.group_names:
```
jnsbck, line 1254: An issue with tracking groups in the nodes dataframe that could arise is that a group name conflicts with  other channel / param names. Could we add a check that ensure groups can only be added if a column of that name does not exist already?

tests/test_groups.py
```diff
@@ -107,3 +108,15 @@ def test_fully_connect_groups_equivalence(SimpleNet):
...
+def test_set_ncomp_changes_groups(SimpleCell):
+    """Test whether groups get updated appropriately after `set_ncomp`."""
+    cell = SimpleCell(3, 4)
+    cell.branch(0).add_to_group("exc")
+    cell.branch(0).set_ncomp(1)
+    assert len(cell.exc.nodes) == 1
+
+    cell.branch(1).add_to_group("exc")
+    cell.branch(0).set_ncomp(2)
+    assert len(cell.exc.nodes) == 6  # 2 from branch(0) and 4 from branch(1).
```
jnsbck, line 122: also might be worth testing if `set_ncomp` correctly raises if ngroups > 1 within module.

jnsbck (changes requested):
I left a few comments. In general, while I see some upsides of this change,  I am not super convinced of having the groups in `nodes`.
- It will make `nodes` more convoluted and add a bunch of boolean columns in some locations (possibly between inserted mechanisms). 
- There will also be no visual distinction of a `Na` group and `Na` channel. The user would have to know that (w.o. looking into `group_names`).
- I am not a big fan of many `x_names` attributes in `Module`. I think dictionaries are more helpful.

Two ideas:
1. replace the old `groups` Dict with a dataframe of booleans
2. attach the group booleans only when you need them for `set_ncomp`.

1. would keep `nodes` clutter-free, while retaining all the functionality we'd need for et_ncomp` to work and makes groups a more distinct object.

Lemme know what you think.

michaeldeistler:
I also thought about this a bit more. Overall, I think I tend to prefer having `groups` as part of `nodes`.

1) It does not force the user to learn another attribute (i.e. the user has to know that they can inspect groups via `.groups`
2) Things are just so much easier if one does not track `groups` separately. E.g for the graph backend, one needs a special case to assign the group indices to the graph. I think this has two major disadvantages:
  - it forces devs to know about the graph backend (and possibly other parts of the software) when anything about the groups is to be changed
  - it is much more error prone because we might just forget to consider groups (as already happened).

In my opinion, these concerns outweigh the disadvantage of cluttering `.nodes` more. All other disadvantages could be fixed I think:
- we could get rid of `group_names` (and of `channel_names`) and replace them with `nodes.columns[nodes.dtypes == bool]`
- have an assert that a group cannot be named as a column in `nodes`.

I still have not fully made up my mind, it's tricky. Happy to zoom!
Michael

jnsbck:
I see the upsides as well, especially with respect needing fewer lines of code. Overall I also don't mind the fact that more things are displayed in nodes. 

> However, we always treat channels and groups the same anyways.

While `comp.group_or_channel_name` works the same, I don't think of groups and channels as the same conceptually. For example `comp.insert(HH())` will add a channel (incl. channel attributes) **and** a group, while `comp.add_to_group("HH")` just adds a group. Now what would happen if the user did sth. like `branch.comp(0).add_to_group("HH")` and `branch.comp(1).insert(HH())`? It's not so clear to me. Does this result in a channel with NaN attrs? Does `branch.HH.set("HH_gNa", 1.0)` change it from NaN to 1.0? Etc.

I can get used to having groups in nodes, but we'd have to make sure that there is some kind of distinction between channels and groups, i.e. check if `group_name in self.channel_names` in `add_to_group` or `channel_name in self.group_names` in `insert` and raise.

PR #601, opened 2025-03-07
Update LIF, implement Izhikevich and rate-based models

michaeldeistler:


jaxley/channels/__init__.py
```diff
@@ -3,5 +3,7 @@
 
 from jaxley.channels.channel import Channel  # isort: skip
 from jaxley.channels.hh import HH
-from jaxley.channels.lif import LIF
+from jaxley.channels.non_capacitive.izhikevich import Izhikevich
```
jnsbck, line 6: I get why you group them in `non_cap`, which I think is fine, but I think splitting this into `channels.rate` and `channels.spike` submodules would arguably be a bit nicer.

jaxley/channels/non_capacitive/izhikevich.py
```diff
@@ -0,0 +1,58 @@
...
+        super().__init__(name)
+        self.channel_params = {
+            f"{self.name}_a": 0.02,
+            f"{self.name}_b": 0.2,
+            f"{self.name}_c": -65.0,
+            f"{self.name}_d": 8,
+        }
+        self.channel_states = {f"{self.name}_u": 0.0}
+        self.current_name = f"{self.name}_izhikevich"
+        warn(
```
jnsbck, line 28: We could also raise `NotImplementetError` using [jax.custom_gradient](https://docs.jax.dev/en/latest/_autosummary/jax.custom_gradient.html) and replace this later with some sort of surrogate. Same goes for the fire model.

tests/test_channels.py
jnsbck, line 1: Sweet!

jnsbck (approved):
Totally love this! Also great tutorial. Only minor suggestions. Good to go otherwise :)

PR #606, opened 2025-03-18
Fix data set indices

### LINKED ISSUE #573: BUG: `.data_set()` function does not work with vectors

michaeldeistler:
Fixes #573

Code by @chaseking

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -893,11 +893,13 @@ class Module(ABC):
         viewed_inds = self._nodes_in_view if is_node_param else self._edges_in_view
         if key in data.columns:
             not_nan = ~data[key].isna()
+            indices = jnp.asarray(viewed_inds[not_nan]).reshape(-1, 1)  # shape (n_comp, 1)
+            val = jnp.broadcast_to(jnp.asarray(val), (indices.shape[0],))  # shape (n_comp,)
             added_param_state = [
                 {
-                    "indices": jnp.asarray(viewed_inds[not_nan]).reshape(-1, 1),  # shape (n_comp, 1)
+                    "indices": indices,
                     "key": key,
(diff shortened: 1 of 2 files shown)


jaxley/modules/base.py
```diff
@@ -876,7 +876,7 @@ def data_set(
         self,
         key: str,
         val: Union[float, jnp.ndarray],
-        param_state: Optional[List[Dict]],
+        param_state: Optional[List[Dict]] = None,
```
michaeldeistler, line 879: I would prefer to not have `None` default here, as it might make the following bug more frequent:
```python
def sim(params):
    pstate = cell.data_set("radius", 2.0)
    pstate = cell.data_set("length", 20.0)
    # ....
```

michaeldeistler (commented):
Awesome, I love that this fix is so minimal!

Three small things:
- please add a line in `CHANGELOG.md` (under a new section v0.9.0) describing the change
- run `black sbi` and `black tests` for formatting
- let's not have `None` as default, see below

Thanks a lot for fixing this!

jnsbck (approved):
Looks great! Thanks a lot for the pr!

michaeldeistler:
Just merged this. Thanks again for the contribution @chaseking!

PR #612, opened 2025-03-27
`morph_connect` and `morph_delete` methods

michaeldeistler:
### Features for editing morphologies

This PR introduces `morph_delete` and `morph_connect`:
```python
from jaxley.io.morph_utils import morph_delete
cell = morph_delete(cell.apical)  # Creates a new cell which has the apical dendrite deleted.
```
```python
from jaxley.io.morph_utils import morph_connect
new_cell = morph_connect(cell.branch(0).loc(0.0), stub.branch(0).loc(0.0))
```

### Integrate radius, don't interpolate

- after this PR, we _integrate_ the radius and do not interpolate. We now have a perfect match with radiuses of NEURON. Because of this (and the graph reader backend as default, see below), I had to adapt the voltages in `tests/jaxley_identical/test_swc.py`.

### Graph backend is now the default SWC reader

`jx.read_swc()` now has `graph` as backend by default. The `SimpleMorphCell` also uses the `graph` backend in tests by default now. To make this work, I had to:
- fix radiuses not getting updated properly after `set_ncomp`
- adapt NEURON <--> Jaxley branch mapping
- fix a couple of voltages because we no longer have the root compartment

### Documentation

This PR also updates the docs:
- add a how-to guide to the website (contains brief and practical steps for specific tasks)
- beautify the documentation landing page.
- beautify the API reference docs.

### Branchpoints

Finally, this PR introduces `module._branchpoints` which also get tracked during `View`.


jnsbck:
Cool, I like it. Should we also add a "getting started" snippet ala (define comp, add mech, add stim, add rec, integrate, plot)?

tests/test_graph.py
```diff
@@ -409,4 +413,110 @@ def integrate():
...
+    )
+
+
+@pytest.mark.parametrize("ncomp", [1, 2])
+def test_morph_edit_swc(ncomp: int):
+    """Check whether we get NaN after having deleted and added things to SWC."""
+    dirname = os.path.dirname(__file__)
+    fname = os.path.join(dirname, "swc_files", "morph_l5pc_with_axon.swc")
+    cell = jx.read_swc(fname, ncomp=ncomp, backend="graph")
+    cell = morph_delete(cell.axon)
```
jnsbck, line 492: should we test repeated delete and what happens when the user adds something twice?

jnsbck: could test morph_attach and then morph_delete gives same thing as before. But might be overkill

jaxley/morphology/morph_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        # be trimmed away `_remove_branch_points_at_tips`, which is run during
+        # `from_graph()`.
+        if (
+            comp_graph.nodes[node]["type"] == "branchpoint"
+            or comp_graph.nodes[node]["comp_index"] not in comps_to_delete
+        ):
+            nodes_to_keep.append(node)
+
+    comp_graph = nx.subgraph(comp_graph, nodes_to_keep)
+    cell = from_graph(comp_graph)
```
jnsbck, line 49: variable name `cell` could be a bit confusing since it should work on any view no?

jnsbck: if it only works for cells, we should assert that there is only a single cell index in nodes or sth.

jaxley/morphology/morph_utils.py
```diff
@@ -0,0 +1,105 @@
+# This file is part of Jaxley, a differentiable neuroscience simulator. Jaxley is
+# licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
+
+import networkx as nx
+import numpy as np
+
+from jaxley.io.graph import connect_graphs, from_graph, to_graph
+
+
+def morph_delete(module_view) -> "Cell":
```
jnsbck, line 10: I love how minimal this is with the new graph backend!

jnsbck (commented):
This is really cool. I like all the new documentation and that the delete/attach turned out to be really simple with the new backend. I still need to look at `io.graph` but got too tired, so I am calling it a day now. Stlil wanted to post the review though. I left a bunch of comments. In general this looks really good. Will take a final look at `io.graph` on Tue, but happy for it to be merged then.

docs/faq/question_05.md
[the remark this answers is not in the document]
michaeldeistler: > FAQ could be a quick way to find how to do things

I would suggest that these things will live in the how-to guide. FAQ will only have high-level guidance such as "what are the units", "what models can I implement", "why should I use Jaxley",...

CONTRIBUTING.md
```diff
@@ -11,8 +13,9 @@ your feature branch to the upstream `Jaxley` ([details](https://docs.github.com/
 
 ### Development environment
 
-Clone [the repo](https://github.com/jaxleyverse/jaxley) and install via `setup.py` using `pip install -e ".[dev]"` (the dev
-flag installs development and testing dependencies).
+Clone [the repo](https://github.com/jaxleyverse/jaxley) and install via `setup.py`
+using `pip install -e ".[dev, doc]"` (the dev flag installs development and testing
```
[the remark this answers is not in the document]
michaeldeistler: I think they should, because it would be great if developers always add an `example` as part of the docstring, such that it can be displayed on the API ref.

jnsbck (approved):
Had another look at `graph`, but no more comments to add. Feel free to merge it, if you feel, the prev. comments are all addressed.

PR #621, opened 2025-04-14
View tracks ._comp_edges

michaeldeistler:
After this PR, `._comp_edges` gets updated in `View`:
```python
cell.branch(1).loc(0.0)._comp_edges
```

This PR also introduces a few special rules for this:
- for `cell.branch(1)` all edges to neighboring branchpoints are included.
- for `cell.branch(1).comp(0)` no edges to branchpoints are included.
- for `cell.branch(1).loc(0.0)` edges to the parent branchpoint are included.
- for `cell.branch(1).loc(1.0)` edges to the child branchpoint are included.

Why do we need these special rules? For modifying a morphology, we need some method to indicate to which end of a branch we want to attach another morphology. Consider the case where all branches have 1 compartment. Then:
`morph_attach(cell.branch(0).comp(0), new_morph)` would not know whether to attach the `new_morph` at the end or at the beginning of the `comp(0)`. This is now possible by simply distinguishing between
- `morph_attach(cell.branch(0).loc(0.0), new_morph)`
- `morph_attach(cell.branch(0).loc(1.0), new_morph)`

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -27,6 +27,7 @@ from jaxley.synapses import Synapse
 from jaxley.utils.cell_utils import (
     _compute_index_of_child,
     _compute_num_children,
+    _get_comp_edges_in_view,
     build_radiuses_from_xyzr,
     compute_axial_conductances,
     compute_levels,
@@ -124,6 +125,8 @@ class Module(ABC):
         self._scope = "local"  # defaults to local scope
         self._nodes_in_view: np.ndarray = None
         self._edges_in_view: np.ndarray = None
+
+        self._comp_edges: pd.DataFrame = pd.DataFrame()
 
         self.edges = pd.DataFrame(
             columns=[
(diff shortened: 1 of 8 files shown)


jaxley/modules/base.py
```diff
@@ -593,7 +614,29 @@ def loc(self, at: Any) -> View:
...
+            comp_edge_condition = "startpoint"
+        elif at == 1.0:
+            comp_edge_condition = "endpoint"
+        else:
+            # For any `0 < at < 1`, we do not add any edges to branchpoints.
+            comp_edge_condition = "source_and_sink"
+        # This could also use `.comp(global_comp_idxs)` instead of
+        # `._at_nodes("comp", global_comp_idxs)`, but this would force us to add
+        # `comp_edge_condition` as an attribute to `.comp()`, which @michaeldeistler
+        # found ugly (because it is user-facing).
```
jnsbck, line 632: +1

jaxley/modules/base.py
```diff
@@ -2792,6 +2864,37 @@ def _set_inds_in_view(
...
+                        ((pre | post) & is_parent).flatten()
+                    ]
+                else:
+                    raise ValueError(
+                        f"comp_edge_condition is {comp_edge_condition}, but must be in "
+                        "{source_and_sink, source_or_sink, startpoint, endpoint}."
+                    )
+                self._comp_edges_in_view = np.intersect1d(
+                    possible_edges_in_view, self._comp_edges_in_view
+                )
```
jnsbck, line 2897: it might make sense to seperate this out into an extra function ala `_set_comp_edges_in_view` for clearity

jnsbck: ...and add a doc string

tests/test_viewing.py
```diff
@@ -621,3 +621,44 @@ def test_view_equals_module(SimpleComp, SimpleBranch):
             )
         ]
     )
+
+
```
jnsbck, line 625: you might also wanna add a line to `test_view_equals_module` to make sure the attr behaves the same for views and modules

jaxley/modules/compartment.py
```diff
@@ -93,3 +93,6 @@ def _init_morph_jax_spsolve(self):
         self._data_inds = data_inds
         self._indices_jax_spsolve = indices
         self._indptr_jax_spsolve = indptr
+
+        # To enable updating `self._comp_edges` during `View`.
+        self._comp_edges_in_view = self._comp_edges.index.to_numpy()
```
jnsbck, line 98: any reason why this cannot be part of Module and needs to be part of Compartment, Branch, Cell, Network?

jnsbck: and is there a reason why comp_edges and comp_edges_in_view have different types? array vs DataFrame?

jnsbck (approved):
I must admit I am not a huge fan of additional kwargs on View, but this is just from my superficial knowledge about the implementation of `morph_attach`, so if you think its necessary or the most elegant solution, lets keep it.

Other than that I left a few comments, but PR looks mostly good to to me and can be merged if CI passes.

jaxley/modules/compartment.py
michaeldeistler: 1) Because it depends on `comp_edges`, which is specific to each type of Module.
2) It's the same for `nodes` and `nodes_in_view` as well as `edges` and `edges_in_view`.

jaxley/modules/base.py
```diff
@@ -515,7 +536,7 @@ def _at_nodes(self, key: str, idx: Any) -> View:
         where = self.nodes[self._scope + f"_{key}_index"].isin(idx)
         inds = self.nodes.index[where].to_numpy()
 
-        view = View(self, nodes=inds)
+        view = View(self, nodes=inds, comp_edge_condition=comp_edge_condition)
```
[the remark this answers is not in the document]
michaeldeistler: I think this could work, but imo it would be much more hacky. I also think that a notion of a branchpoint will be useful for many more things down the road (e.g. for visualization of which end of a branch is `0.0` and which one is `1.0`, or even for recording voltage at a branchpoint), so I would prefer to keep it as it is.

PR #623, opened 2025-04-17
Remove JAX version pin

michaeldeistler:
See https://github.com/jax-ml/jax/issues/26145#issuecomment-2813135240


jnsbck:
This is awesome news!

PR #628, opened 2025-05-01
Fixed minor bug/typo in ChainTransform

kyralianaka:
I found this minor bug in ChainTransform so I'll submit this PR. Not sure if this requires updating the changelog, additional testing and the whole nine yards, but if so I might have to either have someone else do that or get back to it in a few weeks. 
Edit: I updated the changelog but didn't add tests, for my own personal use it works though


michaeldeistler (approved):
Thanks!

PR #641, opened 2025-05-23
Bump version number to 0.9.0

michaeldeistler:


jnsbck (approved):
I am fine with releasing. At some point we should discuss what warrants a major version change like going from 0.8.2 -> 0.9.0 vs. 0.8.3.

michaeldeistler:
The usual rule for version numbering is that new features would lead to the second digit being increased. The third digit gets increased for bugfixes.

PR #643, opened 2025-05-26
Synapse recording indices bug fix

kyralianaka:
The test added in this pull request used to fail because the rec_index came from `self._edges_in_view` (indices from the whole edges dataframe) while jaxedges saves synapse states separately by synapse type (thanks for your investigation @michaeldeistler!). This then means that synapse states returned as the recordings were incorrect if their global edge index exceeded the number of edges of that synapse type. In this case the recording would have mirrored the last synapse state timecourse within range.

People who use multiple synapse types and record synapse states would have been strongly affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.9.1
+
+### 🐛 Bug fixes
+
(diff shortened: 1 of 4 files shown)


michaeldeistler (approved):
Thank you, LGTM!

PR #648, opened 2025-06-06
Pathwise distance between compartments

### LINKED ISSUE #572: Distance function does not calculate path length
Hi @michaeldeistler and jaxley team -- I'm working with a few folks to port the biophysical model implemented in this [paper](https://www.biorxiv.org/content/10.1101/2023.06.02.543490v3) from NEURON to Jaxley.

Crucial to this model is that the channel conductances vary by distance to the Soma, where distance is the path length along the tree. Looking at the source code for [base.py](https://github.com/jaxleyverse/jaxley/blob/e6f7ab096533cebde115caf6a5e5537f353af45d/jaxley/modules/base.py#L1181) it seems like the distance returned by Jaxley is just the euclidean distance. Is this true? If so, I'd like to help remedy this!

A second and related question, is that it's not very clear to me what the right way is to iterate over a cell and calculate the distance from each compartment to the soma. Something that I'd expect to work is roughly:

```
 for branch in cell.branches:
        for c in branch.compartments:
            # get distance to soma
```

but this doesn't work. Would you mind clarifying what the right way to do this is?

Thanks a bunch! Ben

michaeldeistler:
Closes #572 

- deprecate `module.distance()`. It is replaced by `jx.morphology.distance_direct()`
- introduce `jx.morphology.distance_pathwise()`
- the new `distance_pathwise()` allows users to compute pathwise distance.

### Examples

Example 1: The following computes the pathwise distance between the zero-eth soma compartment and all other compartments. It then saves this distance in
`cell.nodes["path_dist_from_soma"]`.
```python
from jaxley.morphology import distance_pathwise

path_dists = distance_pathwise(cell.soma.branch(0).comp(0), cell)
cell.nodes["path_dist_from_soma"] = path_dists
```

Example 2: The following computes the pathwise distance between two compartments.
```python
from jaxley.morphology import distance_pathwise

dist = distance_pathwise(cell.branch(8).comp(2), cell.branch(2).comp(0))
```

Example 3: The following computes the direct (line of sight) distance between the zero-eth soma compartment and all other compartments. It then saves this distance in `cell.nodes["direct_dist_from_soma"]`.

```python
from jaxley.morphology import distance_direct

cell.compute_compartment_centers()  # necessary if you modified branch length.
direct_dists = distance_direct(cell.soma.branch(0).comp(0), cell)
cell.nodes["direct_dist_from_soma"] = direct_dists
```

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1457,10 +1457,21 @@ class Module(ABC):
             for key in list(synapse.synapse_states.keys()):
                 self.base.edges.loc[condition, key] = all_states[key]
 
+    @deprecated(
+        "v0.11.0",
+        (
+            " Instead, please use, e.g., "
+            "`jx.morphology_utils.distance_direct(cell[0, 0], cell[2, 1])`. "
+            "Note that, unlike `cell[0, 0].distance(cell[2, 1]), that "
+            "function returns a list of distances (to all endpoints)."
+        )
+    )
     def distance(self, endpoint: "View") -> float:
         """Return the direct distance between two compartments.
-        This does not compute the pathwise distance (which is currently not
-        implemented).
+        
+        This function computes the direct distance. To compute the pathwise distance,
+        use `distance_pathwise()`.
+
         Args:
             endpoint: The compartment to which to compute the distance to.
         """
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1464,11 +1464,11 @@ class Module(ABC):
             "`jx.morphology_utils.distance(cell[0, 0], cell[2, 1], kind='direct')`. "
             "Note that, unlike `cell[0, 0].distance(cell[2, 1]), that "
             "function returns a list of distances (to all endpoints)."
-        )
+        ),
     )
     def distance(self, endpoint: "View") -> float:
         """Return the direct distance between two compartments.
-        
+
         This function computes the direct distance. To compute the pathwise distance,
         use `distance_pathwise()`.
 
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1461,7 +1461,7 @@ class Module(ABC):
         "v0.11.0",
         (
             " Instead, please use, e.g., "
-            "`jx.morphology_utils.distance_direct(cell[0, 0], cell[2, 1])`. "
+            "`jx.morphology_utils.distance(cell[0, 0], cell[2, 1], kind='direct')`. "
             "Note that, unlike `cell[0, 0].distance(cell[2, 1]), that "
             "function returns a list of distances (to all endpoints)."
         )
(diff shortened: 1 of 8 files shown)


jaxley/morphology/distance_utils.py
```diff
@@ -0,0 +1,109 @@
...
+import jax.numpy as jnp
+import networkx as nx
+
+from jaxley.modules.base import to_graph
+
+
+def distance(
+    startpoint: "View",
+    endpoints: Union["Branch", "Cell", "View"],
+    kind: str = "pathwise",
```
jnsbck: I slightly prefer having separate functions, rather than a kwarg, but fine with either.

jaxley/morphology/distance_utils.py
```diff
@@ -0,0 +1,109 @@
...
+            length = sum(graph.nodes[n]["length"] for n in path)
+            length -= 0.5 * graph.nodes[path[0]]["length"]  # subtract half root
+            length -= 0.5 * graph.nodes[path[-1]]["length"]  # subtract half target
+            return length
+
+        # Use Dijkstra to minimum adjusted length from root to all endpoint nodes.
+        lengths = []
+        for endpoint in endpoint_inds:
+            path = nx.shortest_path(graph, root, endpoint)
+            lengths.append(custom_path_length(graph, path))
```
jnsbck: You could use the [nx dijkstra's](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.weighted.single_source_dijkstra.html) implementation as well. Something  like this should work.

```python
def edge_weight(u, v, d):
    _u = graph.nodes[u]
    _v = graph.nodes[v]
     return ((_u["x"] - _v["x"])**2 + (_u["y"] - _v["y"])**2 + (_u["z"] - _v["z"])**2)**0.5

length = nx.single_source_dijkstra_path_length(graph, startpoint, endpoint, weight=edge_weight)
```

jnsbck (approved):
Looks good to me. I like that this lives in morphology. 

I would prefer to move the pathwise distance to `nx` implementation of dikstra, but otherwise looking good.

jaxley/modules/base.py
```diff
@@ -1457,10 +1457,21 @@ def write_trainables(self, trainable_params: List[Dict[str, jnp.ndarray]]):
             for key in list(synapse.synapse_states.keys()):
                 self.base.edges.loc[condition, key] = all_states[key]
 
+    @deprecated(
+        "v0.11.0",
```
jnsbck: just noticed, I think you have to specify this without _v_.

jaxley/morphology/distance_utils.py
michaeldeistler: I was really unsure about this, so if you prefer two functions then I will split them :)

jaxley/morphology/distance_utils.py
michaeldeistler: This worked, but with two modifications:
1) use `nx.dijkstra_path_length` because `nx.single_source_dijkstra_path_length` computes the length from a single start node to _all_ end nodes.
2) edge weight should be: `(graph.nodes[u]["length"] + graph.nodes[v]["length"]) / 2`, because your implementation computes the euclidean (direct) distance between two compartments, not the pathwise dist.

PR #660, opened 2025-06-13
Refactor of SWC Reader.

### LINKED ISSUE #596: Add scaling factor to swc reader
As it turns out that not all swc-like files list x, y, z, and r in um, we talked about adding an optional scaling factor argument to `read_swc`. This would then likely be one line in `swc_to_jaxley` that does the conversion of the relevant columns of `content`. Let me know if this sounds alright to you all and then I can write it and use it.

### LINKED ISSUE #741: `build_compartment_graph` doesn't follow min_radius when computing electrical properties
Hi Jaxley team -- congrats on the paper! Really nice to see that it's finally out.

Anyway, I noticed an issue with the way that `build_compartment_graph` handles the `min_radius` argument. Right now, the `morph_attrs_from_xyzr` uses the values of the radii _before_ they are clipped to the value of `min_radius.` This means that effectively the min_radius argument is ignored when setting the properties of the cell. 

I noticed this when I was trying to load a (weird) SWC file where the radii were all zeros. I thought that I would be able to get a working model by passing `min_radius`, but the resulting values like resitive_load_in, resistive_load_out were all NaN.

Suggested fix: Before calling swc_radius, swc_area, swc_volume, or swc_resistive_load in morph_attrs_from_xyzr, clip the xyzr radii to min_radius (when provided). I'm happy to submit a PR for this if helpful.

jnsbck:
This rewrite of the graph-based SWC reader took ages, but I had a few quiet days and finalized it. It adds a clearer factorization of the graph pipeline and adds a NEURON backend (as an alternative and also for better ground truthing). This also affects a few other things like `set_ncomp`, which is now faster.

The import now works roughly as follows:
```python
from jaxley.io.graph import swc_to_pandas, swc_to_nx, build_compartment_graph, from_graph, to_graph

swc        = swc_to_pandas("morph.swc")                  # DataFrame
swc_graph  = swc_to_nx(swc)                              # node per traced point
comp_graph = build_compartment_graph(swc_graph, ncomp=4) # node per compartment
cell       = from_graph(comp_graph)                      # jx.Cell

graph = to_graph(cell)                                   # comps + branchpoints + tips
graph = to_graph(cell, channels=False, synapses=False)   # bare morphology
```

`jx.read_swc(...)` still works as before but `ncomp` can now be passed as a callable for per-branch discretization (e.g. d-lambda) at import time, and `cell.branch("all").set_ncomp([3, 5, 1, ...])` can now take one value per branch.

`jx.read_swc(..., backend="neuron")` is a new backend that imports through NEURON's `Import3d` and
basically reproduces NEURON's imports exactly (including the bugs lol). 

The new reader is also much faster! On the l5pc (324 branches), I get the following on my machine:

| | old | new | |
|---|---|---|---|
| `read_swc(ncomp=1)` | 2504 ms | 377 ms | 6.6× |
| `read_swc(ncomp=4)` | 6679 ms | 961 ms | 6.9× |
| `set_ncomp` 1 → 4, whole cell | 829 s | 1.6 s | ~500× |

The PR (according to Claude) also reduces linecount by ~10% for the new pipeline.

More fixes:
- Compartment centers are now the point at half the compartment's arc length, not the midpoint
  of the chord between its ends. This was previously off for long compartments of curved branches.
- Branch tracing no longer depends on SWC node order
- Better and more stable tests against NEURON (compartments are now matched by the coords of their centers instead of per index for example)
- `_init_solvers` now builds its DHS graph from `_comp_edges` so  `jaxley.modules` has no more dependency on `jaxley.io`. I also moved `to_graph`

Some more things to note:
- `jaxley/io/graph.py` is the new reader core. `jaxley/io/neuron.py` is the new
  backend and `jaxley/io/legacy.py` all the legacy graph reader code + dependencies. It should be self contained and is only tested via `tests/test_graph_legacy.py`. Deprecating it should be straight forward should we want to do this at some point when the new reader has proven reliable.
- `jaxley/modules/base.py`: `set_ncomp` now does basically only `_rediscretize` + `_compartmentalize_branch`

TODOs before merge
- [ ] Regenerate the hardcoded test values, since the new reader is marginally different. The importent test that is passed though is the comparison against NEURON on voltage.
- [ ] Update the API docs
- [ ] Update the tutorials that use the old graph API.
- [ ] Delete `jaxley/io/legacy.py`, `tests/test_graph_legacy.py` and the `"legacy"` branch. (at some point)

Follow-ups, not in this PR

- Route `set_ncomp` through the graph pipeline. Currently `_compartmentalize_branch` recomputes from `xyzr`
  what `build_compartment_graph` already does. Can be done via `compartmentalize()` too. Currently only pinned by tests.
- `_build_module` hand-assigns module attributes that duplicate `Module.insert()` and
  `_append_multiple_synapses()`, so could drift silently if those change.


huangziwei:
btw, have you ever tried [igraph](https://python.igraph.org/en/stable/)? Switching to it from `nx` gave me at least 2x~3x boost when computing shortest path distance. the documentation is not as good as `nx` though, but llm should be able to to work out most of the common usages.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.11.0 (pre-release)
+
+### 🛠️ Internal updates
+- refactor of the graph import. Now more better, more faster and has fewer lines (#660, @jnsbck)
```
michaeldeistler: Now better, faster, and has fewer lines.

jaxley/io/tmp.py
```diff
@@ -0,0 +1,668 @@
+from __future__ import annotations
```
michaeldeistler: tmp.py?

Should be removed i think.

michaeldeistler: Or renamed? I am confused...

jaxley/io/tmp.py
```diff
@@ -0,0 +1,668 @@
...
+
+def compute_xyz(
+    G: nx.DiGraph,
+    length: float = 1.0,
+    spread: float = np.pi / 8,
+    spread_decay: float = 0.9,
+    twist: float = 0.0,
+    xy_only: bool = True,
+) -> Dict[int, tuple[float, float, float]]:
+    """Compute xyz coordinates for a tree-like appearance of a networkX graph in 2D or 3D.
```
michaeldeistler: this line feels to long?

jaxley/io/tmp.py
```diff
@@ -0,0 +1,668 @@
...
+
+    Handles branches implicitly since nodes in a branch have 1 child.
+
+    Args:
+        G: The Graph to compute node xyz coordinates for.
```
michaeldeistler: let's name graphs `graph`, not `G` (linters complain about `G`)

jaxley/io/tmp.py
```diff
@@ -0,0 +1,668 @@
...
+        xy_only: Whether to only compute the xy coordinates and fix the z-coordinate.
+
+    Returns:
+        A dictionary mapping node indices to xyz coordinates.
+    """
+    # TODO: Replace compute_xyz with this or vice versa, redundant!
+    root = next(n for n, d in G.in_degree() if d == 0)
+    pos = {root: (0.0, 0.0, 0.0)}
+
+    def recurse(node, depth=1, theta=0.0, phi=np.pi / 2):
```
michaeldeistler: docstring, and please make the function name more explicit.

jaxley/io/tmp.py
```diff
@@ -0,0 +1,668 @@
...
+            recurse(child, depth + 1, th, ph)
+
+    recurse(root, theta=0.0, phi=np.pi / 2)
+    return pos
+
+
+def pandas_to_nx(
+    node_attrs: pd.DataFrame, edge_attrs: pd.DataFrame, global_attrs: pd.Series
+) -> nx.DiGraph:
+    """Convert node_attrs, edge_attrs and global_attrs from pandas datatypes to a NetworkX DiGraph.
```
michaeldeistler: line length.

michaeldeistler (changes requested):
Hi @jnsbck,

I started reviewing this but I am really confused---why is all code in `tmp.py`? Why are zero lines removed? I think there is some stuff missing. Please add it and request another review. Thanks!

CHANGELOG.md
jnsbck: ...it is grammatically questionable, I admit, but I wanted the "more"s in there :D

jaxley/io/tmp.py
jnsbck: Yea, thats why I wrote:
> I have not started integrating it yet and kept it seperate for now. Feedback is greatly appreciated.
I thought you might wanna take a look, while it has not been integrated yet, in case you request mayor changes, and to make it easier to step through. :)

CHANGELOG.md
michaeldeistler: I am fine with wording this strongly, and also with something being grammatically questionable, but I would prefer to not have things that are straight up grammatically wrong :)

michaeldeistler (commented):
Overall this looks awesome, I love how short and fast it is.

A few things:

1) Let's make sure that graphs are always named explicitly (I see you already do this in later parts of the code of dev.ipynb, but let's also do it consistently in all python modules). I am not a fan of something like:
```python
G = swc_to_nx("../jaxley/tests/swc_files/morph_interrupted_soma.swc")
G = compartmentalize(G, ncomp=1)
```
and I would prefer
```python
swc_graph = swc_to_nx("../jaxley/tests/swc_files/morph_interrupted_soma.swc")
comp_graph = compartmentalize(swc_graph, ncomp=1)
```

2) I actually get much larger speedups than you :) Between 3x and 15x.

3) You rename quite some functions, e.g. `swc_to_nx`, `compartmentalize`. I don't think I have strong opinions, but let's carefully think about naming if we want to change them (again).

4) Let's have as many code comments on what is happening as possible. I find these parts of the toolbox to be very difficult to follow purely based on code (in part because I do not know nx very well).

5) Before making things "nice", I would recommend you to test whether all SWC tests pass (if you have not done so yet).

CHANGELOG.md
jnsbck: I'll change it ;)

PR #662, opened 2025-06-17
Fixup for computing membrane area for SWC files

### LINKED ISSUE #650: Jaxley and NEURON voltage trace mismatch
Hello Jaxley community:

In the Jaxley paper it claims Jaxley's simulation matches that of NEURON's. However I saw a mismatch with NEURON and Jaxley in the following conditions:

Jaxley: GPU, dhs; NEURON: CPU, default solvers
Mismatch with same SWC file and ephys parameter input (simulated only one cell, specified capacitance, eLeak, gLeak, axial resistivity; only leak mechanism was inserted).

Multiple SWC files were tested, including SWC files containing 3, 50, 1000, 10000, and 33000 lines. I saw mismatch in all conditions. (attached is for 10000 lines)

Image

I am wondering if it is possible that the voltage solvers are difference and caused the mismatch, or are there different ways Jaxley and NEURON handles morphology (especially radius and length).

I considered numerical instability but ODEs for neuron simulation should be benign and diffusive (not stiff) and I don't think numerical issues would be relevant.

michaeldeistler:
Fixes #650 

Here is a minimal SWC file that reproduces the issue in #650:
```
1 0 0 0 0 2.0 -1
2 0 0 1 0 0.2 1
```

In `NEURON` (and in `Jaxley`), this file gets translated into a _point neuron_ (i.e., single compartment neuron). However, the `.area` of that neuron get computed differently:
- `Jaxley` used `area = 2 * pi * r * l`, where `r` gets integrated with trapezoidal integration
- `NEURON` uses a truncated cone approximation to compute the area. As such, the `seg.area()` of `NEURON` only corresponds to `2 * pi * r * l` if the radius within a compartment is constant.

The behavior is similar for the axial resistivity.

This PR fixes the behavior of `Jaxley` and matches it to how `NEURON` does it. I find it quite surprising that we had not caught this previously, but it does actually occurs quite rarely: only if a single compartment has different `r`, and in fact the difference is only noticable if `r` is _quite_ different. In my experiments, it is only visually apparent if the soma did not get assigned a separate `type_id` (and is therefore lumped together with dendritic traced points, leading to a large variation in radius within a compartment).


jaxley/io/swc.py
```diff
@@ -358,81 +353,6 @@ def swc_to_jaxley(
     return parents, pathlengths, radius_fns, types, all_coords_of_branches
 
 
-def read_swc_custom(
```
jnsbck, line 361: If we are deprecating this, we should make sure to delete all the related code that will not be used anymore, i.e. `swc_to_jaxley`, `_split_long_branches` etc. (incl. in `cell_utils`)

jnsbck: and also remove the `backend` kwarg

jaxley/modules/base.py
```diff
@@ -1054,16 +1057,6 @@ def _init_solver_jaxley_dhs_solve(
...
-        """Given radius, length, r_a, compute the axial coupling conductances.
-
-        If ion diffusion was activated by the user (with `cell.diffuse()`) then this
-        function also compute the axial conductances for every ion.
-        """
-        return compute_axial_conductances(
-            self._comp_edges, params, self.diffusion_states
-        )
-
     def set(self, key: str, val: Union[float, jnp.ndarray]):
```
jnsbck, line 1060: naive question. What happens if the user makes "length" and "area" trainable"? Should we add a warning to make trainable?

jaxley/modules/base.py
```diff
@@ -1087,7 +1080,39 @@ def set(self, key: str, val: Union[float, jnp.ndarray]):
...
+
+            # When the key is `radius` or `length`, we also have to update the
+            # membrane surface area. In principle, we could also do this on the fly
+            # when a simulation is started, but computing the membrane area for
+            # SWC-traced neurons can be computationally expensive.
+            if key in ["radius", "length"]:
+                # Add an additional warning if the neuron was read from SWC.
+                xyzr = np.concatenate(self.base.xyzr)
+                xyzr_is_available = np.invert(np.any(np.isnan(xyzr[:, 3])))
+                if xyzr_is_available:
```
jnsbck, line 1094: doesnt `compute_xyz` also populate `xyzr`? So the warning would also be thrown in that case.

jaxley/modules/base.py
```diff
@@ -1619,9 +1674,27 @@ def get_all_parameters(self, pstate: List[Dict]) -> Dict[str, jnp.ndarray]:
                 # `.set()` to work. This is done with `[:, None]`.
                 params[key] = params[key].at[inds].set(set_param[:, None])
 
+            # If radius and length are updated by the pstate, then we have to also
```
jnsbck, line 1691: didnt look, but do we test for this case?

jaxley/modules/branch.py
```diff
@@ -95,6 +95,7 @@ def _init_comp_graph(self):
             {
                 "source": list(range(self.ncomp - 1)) + list(range(1, self.ncomp)),
                 "sink": list(range(1, self.ncomp)) + list(range(self.ncomp - 1)),
+                "ordered": list([1] * (self.ncomp - 1)) + list([0] * (self.ncomp - 1)),
```
jnsbck, line 98: What does ordered do exactly?

tests/swc_files/morph_variable_radiuses_multi_branch.swc
```diff
@@ -0,0 +1,26 @@
+# This file was written by @michaeldeistler. It contains three branches, each with a
+# slightly varying radius.
```
jnsbck: Should we link the issue here for context?

jaxley/modules/base.py
michaeldeistler: Making `length` and `area` trainable would make the code depend on the order in which they are run (same for `.set()`).

Yes, I think a warning is a good idea.

jaxley/modules/base.py
michaeldeistler: It does not populate the radius, which is `xyzr[:, 3]`.

jaxley/modules/branch.py
michaeldeistler: It indicates whether this compartment edge is along the direction of the xyzr coords.

Issue #676, opened 2025-07-04
Bug in set_ncomp


manuelgloeckler:
## Description

Here is a simple, minimal code example  on jaxley 0.9.0:

```python
import jaxley as jx
cell = jx.Cell([jx.Branch(), jx.Branch()], [-1, 0])
cell.branch(0).set_ncomp(5, initialize=False)    # Also fails with initialize=True
cell.branch(1).set_ncomp(10, initialize=False)  # Also fails with initialize=True
cell.initialize()
print(cell.branch(0).ncomp) # -> 10 != 5         # Always return the max ncomp
print(len(cell.branch(0).nodes)) # -> 5 correct
```

So this pattern is e.g., used in [here](https://jaxley.readthedocs.io/en/latest/how_to_guide/set_ncomp.html). But this might not be directly affected as it sets the nodes table correctly. But the ncomp attribute will be set incorrectly for all branches! This can have an adverse effect downstream, e.g., recreating a Cell will fail (just because n_comp is false, replacing it with len(nodes) fixes it).

```python
branches = list(cell.branches)
parents = [p for b in branches for p in b.comb_parents]
new_cell =jx.Cell(branches, parents) # Fails because incorrect ncomp attribute which is used __init__
```

In general, anything that uses `n_comp` per branch will be affected, if not all compartments have the same `n_comp` (the soma is excluded here).

## Problem

Not sure. On a cell level, the max ncomp is used, which is correct and makes sense. But somehow this affects/propages to the other branches.

Should't it be the case the `ncomp == len(nodes)` for all cases?

jnsbck:
Also seems to happen on `main`. The problem can presumably be found in these 2 lines: 

https://github.com/jaxleyverse/jaxley/blob/d85905c72b616515b52695cda54ed1c8f162501d/jaxley/modules/base.py#L3073
https://github.com/jaxleyverse/jaxley/blob/d85905c72b616515b52695cda54ed1c8f162501d/jaxley/modules/base.py#L3089

I.e. when viewing a `Module` via `View` the max ncomp in view does not get updated with respect to the view correctly.
Changing L3089 to `self.ncomp = max(self.nodes.grouby("branch_index").count())` might fix this.

manuelgloeckler:
Mhh, definitely the place where it goes wrong, thanks! Not sure about `max(self.nodes.grouby("branch_index").count())`. This will correctly list the num_compartments, but then return the maximum compartment again (which is wrong).

[...]

PR #677, opened 2025-07-04
fix ncomp

manuelgloeckler:
This addresses #676.


manuelgloeckler:
So it passes the current test suite. Although I'm not sure if the places where this would take effect are covered by the current tests.

I could add something like the toy example #677 to the tests (but otherwise not sure where to test).

michaeldeistler (approved):
Thanks! Looks good from my side, but let's wait for Jonas' input.

PR #691, opened 2025-07-11
Bugfix for solver speed for networks on GPU

michaeldeistler:
Only for networks of multicompartment neurons (not point neurons), and only on GPU, our solver has been extremely suboptimal since `Jaxley` version `0.9.0`, following [this PR](https://github.com/jaxleyverse/jaxley/pull/625).

This is due to a bug which caused the voltages of cells to be processed sequentially, instead of in parallel (If they are multicompartment). This PR provides a minimal fix for this.

### API changes

There are no breaking changes for the API. However, this PR introduces a new keyword argument to the `Network`:
```python
net = jx.Network([cell], vectorize_cells=True)
```
When set to `True`, we automatically use the gpu-based solver and we parallelize across cells. This is the default on GPU. The default on CPU is `False` because it optimizes compile time (and barely affects runtime on CPU).

### Customization

To run the GPU solver _for networks_ on CPU, you have to do:
```python
cell._init_solver_jaxley_dhs_solve(allowed_nodes_per_level=16)
net = jx.Network([cell], vectorize_cells=True)
v = jx.integrate(net)
```

To run the CPU solver _for networks_ on GPU, do:
```python
cell._init_solver_jaxley_dhs_solve(allowed_nodes_per_level=1)
net = jx.Network([cell], vectorize_cells=False)
v = jx.integrate(net)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # 0.11.2 (pre-release)
 
+### 🐛 Bug fixes
+
+- Bugfix for `Network`s on `GPU`: since `v0.9.0`, networks had been very slow on GPU
+because the voltage equations of cells had been processed in sequence, not in parallel.
+This is now solved, giving a large speed-up for networks consisting of many cells (#691,
+@michaeldeistler, thanks to @VENOM314 for reporting)
+
 ### 📚 Documentation
 
-- Remove all content from the old documentation website (#689, @michaeldeistler)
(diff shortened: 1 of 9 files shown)


jnsbck (approved):
Looks good. We could think about using env variables instead of kwargs, for all of these debug options, i.e. `DEVICE="CPU" VECTORIZE_CELLS=1 python3 run_script.py`. I think this might make the API a bit cleaner, since 99% of users are never going to use them.

Issue #469, opened 2024-10-24
Better typing


manuelgloeckler:
Currently, we do not follow good typing practices. Official suggestion are given [here](https://jax.readthedocs.io/en/latest/jax.typing.html), which more or less suggest the following rules
```python
import jax.numpy as jnp
from jax import Array
from jax.typing import ArrayLike
# Type inputs as ArrayLike (floats, np.array, jnp.array...)
# Type output as Array
def f(x: ArrayLike) -> Array:
     return jnp.pow(x,2)
```

Currently we often use `jnp.ndarray` which is an alias for `Array`.

PR #693, opened 2025-07-12
Follow Jax Typing Practices

alexpejovic:
Fixes/implements #469 

Note that additional non-jax typing changed were made, however these changes were only made on the same lines where the jax related typing changes were made. The changes are mostly using Python native types (list, dict, tuple, etc.) instead of types from the typing package (List, Dict, Tuple, etc.). Using Python native types has been standard since Python 3.9, so I hope these additional changes are also okay, since Jaxley claims to require Python>=3.10 anyway.

### DIFF
--- a/jaxley/channels/channel.py
+++ b/jaxley/channels/channel.py
@@ -82,9 +82,7 @@ class Channel:
         }
         return self
 
-    def update_states(
-        self, states, dt, v, params
-    ) -> tuple[Array, tuple[Array, Array]]:
+    def update_states(self, states, dt, v, params) -> tuple[Array, tuple[Array, Array]]:
         """Return the updated states."""
         raise NotImplementedError
 
--- a/jaxley/channels/channel.py
+++ b/jaxley/channels/channel.py
@@ -5,7 +5,8 @@ from abc import ABC, abstractmethod
 from typing import Dict, Optional, Tuple
 from warnings import warn
 
-import jax.numpy as jnp
+from jax import Array
+from jax.typing import ArrayLike
 
 
 class Channel:
@@ -83,12 +84,12 @@ class Channel:
 
     def update_states(
         self, states, dt, v, params
-    ) -> Tuple[jnp.ndarray, Tuple[jnp.ndarray, jnp.ndarray]]:
+    ) -> tuple[Array, tuple[Array, Array]]:
         """Return the updated states."""
         raise NotImplementedError
 
     def compute_current(
-        self, states: Dict[str, jnp.ndarray], v, params: Dict[str, jnp.ndarray]
+        self, states: dict[str, ArrayLike], v: float, params: dict[str, ArrayLike]
     ):
         """Given channel states and voltage, return the current through the channel.
 
@@ -104,9 +105,9 @@ class Channel:
 
     def init_state(
         self,
-        states: Dict[str, jnp.ndarray],
-        v: jnp.ndarray,
-        params: Dict[str, jnp.ndarray],
+        states: dict[str, ArrayLike],
+        v: ArrayLike,
+        params: dict[str, ArrayLike],
         delta_t: float,
     ):
         """Initialize states of channel."""
--- a/jaxley/channels/channel.py
+++ b/jaxley/channels/channel.py
@@ -1,6 +1,5 @@
 # This file is part of Jaxley, a differentiable neuroscience simulator. Jaxley is
 # licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
-
 from abc import ABC, abstractmethod
 from typing import Dict, Optional, Tuple
 from warnings import warn
@@ -87,7 +86,7 @@ class Channel:
         raise NotImplementedError
 
     def compute_current(
-        self, states: dict[str, ArrayLike], v: float, params: dict[str, ArrayLike]
+        self, states: dict[str, Array], v: float, params: dict[str, Array]
     ):
         """Given channel states and voltage, return the current through the channel.
 
--- a/jaxley/channels/hh.py
+++ b/jaxley/channels/hh.py
@@ -35,10 +35,10 @@ class HH(Channel):
 
     def update_states(
         self,
-        states: dict[str, ArrayLike],
+        states: dict[str, Array],
         dt: float,
         v: float,
-        params: dict[str, ArrayLike],
+        params: dict[str, Array],
     ) -> dict[str, Array]:
         """Return updated HH channel state."""
         prefix = self._name
@@ -49,7 +49,7 @@ class HH(Channel):
         return {f"{prefix}_m": new_m, f"{prefix}_h": new_h, f"{prefix}_n": new_n}
 
     def compute_current(
-        self, states: dict[str, ArrayLike], v: float, params: dict[str, ArrayLike]
+        self, states: dict[str, Array], v: float, params: dict[str, Array]
     ) -> float:
         """Return current through HH channels."""
         prefix = self._name
--- a/jaxley/channels/hh.py
+++ b/jaxley/channels/hh.py
@@ -1,9 +1,10 @@
 # This file is part of Jaxley, a differentiable neuroscience simulator. Jaxley is
 # licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
-
-from typing import Dict, Optional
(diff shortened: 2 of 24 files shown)


jaxley/channels/channel.py
```diff
@@ -81,14 +82,12 @@ def change_name(self, new_name: str):
...
-    def update_states(
-        self, states, dt, v, params
-    ) -> Tuple[jnp.ndarray, Tuple[jnp.ndarray, jnp.ndarray]]:
+    def update_states(self, states, dt, v, params) -> tuple[Array, tuple[Array, Array]]:
         """Return the updated states."""
         raise NotImplementedError
 
     def compute_current(
-        self, states: Dict[str, jnp.ndarray], v, params: Dict[str, jnp.ndarray]
+        self, states: dict[str, ArrayLike], v: float, params: dict[str, ArrayLike]
```
michaeldeistler: I think all of them should be `Array` here, otherwise `jit` or `grad` breaks.

jaxley/channels/channel.py
```diff
@@ -104,9 +103,9 @@ def compute_current(
 
     def init_state(
         self,
-        states: Dict[str, jnp.ndarray],
-        v: jnp.ndarray,
-        params: Dict[str, jnp.ndarray],
+        states: dict[str, ArrayLike],
+        v: ArrayLike,
+        params: dict[str, ArrayLike],
```
michaeldeistler, line 107: In `init_state`, `ArrayLike` is fine---this never gets compiled.

jaxley/channels/hh.py
```diff
@@ -34,11 +35,11 @@ def __init__(self, name: Optional[str] = None):
...
         self,
-        states: Dict[str, jnp.ndarray],
-        dt,
-        v,
-        params: Dict[str, jnp.ndarray],
-    ):
+        states: dict[str, ArrayLike],
+        dt: float,
+        v: float,
+        params: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2506,12 +2507,12 @@ def step(
 
     def _step_channels(
         self,
-        states: Dict[str, jnp.ndarray],
+        states: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2506,12 +2507,12 @@ def step(
...
         delta_t: float,
         channels: List[Channel],
         channel_nodes: pd.DataFrame,
-        params: Dict[str, jnp.ndarray],
-    ) -> Tuple[Dict[str, jnp.ndarray], Tuple[jnp.ndarray, jnp.ndarray]]:
+        params: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2527,8 +2528,8 @@ def _step_channels_state(
         delta_t,
         channels: List[Channel],
         channel_nodes: pd.DataFrame,
-        params: Dict[str, jnp.ndarray],
-    ) -> Dict[str, jnp.ndarray]:
+        params: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2565,12 +2566,12 @@ def _step_channels_state(
 
     def _channel_currents(
         self,
-        states: Dict[str, jnp.ndarray],
+        states: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2565,12 +2566,12 @@ def _step_channels_state(
...
         delta_t: float,
         channels: List[Channel],
         channel_nodes: pd.DataFrame,
-        params: Dict[str, jnp.ndarray],
-    ) -> Tuple[Dict[str, jnp.ndarray], Tuple[jnp.ndarray, jnp.ndarray]]:
+        params: dict[str, ArrayLike],
```
michaeldeistler: `Array`

jaxley/modules/base.py
```diff
@@ -2629,12 +2630,12 @@ def _channel_currents(
 
     def _channel_current_components(
         self,
-        modified_state: jnp.ndarray,
-        states: Dict[str, jnp.ndarray],
+        modified_state: ArrayLike,
```
michaeldeistler: same here, `modified_state`, `states`, and `params` are all `Array`.

jaxley/modules/base.py
```diff
@@ -2671,12 +2672,12 @@ def _channel_current_components(
 
     def _step_synapse(
         self,
-        u: Dict[str, jnp.ndarray],
-        syn_channels: List[Channel],
-        params: Dict[str, jnp.ndarray],
+        u: dict[str, ArrayLike],
```
michaeldeistler: And same here.

jaxley/optimize/optimizer.py
```diff
@@ -12,8 +13,8 @@ class TypeOptimizer:
     def __init__(
         self,
         optimizer: Callable,
-        optimizer_args: Dict[str, Any],
-        opt_params: List[Dict[str, jnp.ndarray]],
+        optimizer_args: dict[str, Any],
+        opt_params: list[dict[str, ArrayLike]],
```
michaeldeistler: `Array`

jaxley/optimize/optimizer.py
```diff
@@ -56,7 +57,7 @@ def __init__(
             optimizer = self.base_optimizer(optimizer_args[name])
             self.optimizers.append({name: optimizer})
 
-    def init(self, opt_params: List[Dict[str, jnp.ndarray]]) -> List:
+    def init(self, opt_params: list[dict[str, ArrayLike]]) -> list:
```
michaeldeistler: `Array`

jaxley/optimize/optimizer.py
```diff
@@ -65,7 +66,7 @@ def init(self, opt_params: List[Dict[str, jnp.ndarray]]) -> List:
             opt_states.append(opt_state)
         return opt_states
 
-    def update(self, gradient: jnp.ndarray, opt_state: List) -> Tuple[List, List]:
+    def update(self, gradient: ArrayLike, opt_state: list) -> tuple[list, list]:
```
michaeldeistler: `Array`

jaxley/pumps/ca_pump.py
```diff
@@ -29,19 +28,19 @@ def __init__(self, name: Optional[str] = None):
 
     def update_states(
         self,
-        states: Dict[str, jnp.ndarray],
+        states: dict[str, ArrayLike],
```
michaeldeistler: Same as above: in `Channel`s and `Pump`s, all states and params should be `Array`

michaeldeistler (changes requested):
Thank you for tackling this, this is awesome!

Two things:
1) In several cases (e.g., channels), using a `np.ndarray` would break `Jaxley` because it automatically jit-compiles the integration loop. As such, I would prefer to use `Array` in these cases (even if the function itself works with `ArrayLike`. What do you think?

2) Please add a line to the `CHANGELOG.md`.

Thanks again!
Michael

alexpejovic:
Input parameter types I changed to Array without explicit request,

alexpejovic:
Feel free to tell me I'm very wrong, but I imagine an ideal scenario would be where all public functions that specify Array as a parameter type would enforce it, for example by throwing an error when a numpy array is passed through it. This would be a pain to implement throughout the whole code base, but if there is explicit separation of public and private functions/classes, then you could only implement such functionality only on public functions, and let a static type checker like mypy handle the private functions. Those are just some ideas I'm throwing out there.

michaeldeistler (approved):
I like the updated version, thank you!

I think the main source of error for users is that `Channel`s must be fully written in `jax.numpy`, not `numpy`. Having checks for this could be useful, but let us leave this for a future PR.

As for using `mypy`: I am not generally opposed to using it, but I think it would require quite a bit of work to get started.

Either way: Please add a line to the `CHANGELOG.md` (new section `0.11.3 (pre-release)`) such that CI tests are passing :)

Apart from this the PR is good to go from my side!

PR #702, opened 2025-08-22
Bugfix for networks that combine point neurons and morph neurons

michaeldeistler:


jaxley/modules/network.py
```diff
@@ -161,11 +161,17 @@ def _init_comp_graph(self):
...
+            if len(branchpoints) > 0:
+                self._branchpoints.append(branchpoints)
 
             offset += cell._n_nodes  # Compartment-offset.
         self._comp_edges = pd.concat(self._comp_edges, ignore_index=True)
-        self._branchpoints = pd.concat(self._branchpoints)
+        if len(self._branchpoints) > 0:
+            self._branchpoints = pd.concat(self._branchpoints)
+        else:
+            self._branchpoints: pd.DataFrame = pd.DataFrame(columns=["x", "y", "z"])
```
michaeldeistler, line 174: The changes in this file are unrelated---they only get rid of a pandas warning.

jaxley/utils/solver_utils.py
```diff
@@ -123,7 +123,11 @@ def dhs_permutation_indices(
         new_node_order.append(
             [mapping_dict[int(n[0])], mapping_dict[int(n[1])], int(n[2])]
         )
-    new_node_order = jnp.asarray(new_node_order)
+    # We need `.reshape(-1, 3)` because `new_node_order` is empty for point-neurons.
+    # However, for networks, we still want to be able to concatenate this empty array
+    # with `(N, 3)` arrays (e.g., from morph detailed neurons). The line below ensures
+    # that `new_node_order` is of shape (0, 3) if it is empty.
+    new_node_order = jnp.asarray(new_node_order, dtype=int).reshape(-1, 3)
```
jnsbck, line 130: if node_order of a single comp is of shape (0,3) what has shape (1,3) then, shouldnt single comp be (1,3)?

tests/jaxley_identical/test_basic_modules.py
```diff
@@ -429,3 +429,80 @@ def test_complex_net(voltage_solver, SimpleCell):
     max_error = np.max(np.abs(voltages[:, ::40] - voltages_300724))
     tolerance = 1e-8
     assert max_error <= tolerance, f"Error is {max_error} > {tolerance}"
+
+
+@pytest.mark.parametrize(
```
jnsbck, line 434: This is quite a big test, shouldnt testing for instantiation not be sufficient? Alternatively, we could just change testing net to include a simple cell, then we dont have to run this twice.

jaxley/utils/solver_utils.py
michaeldeistler: no, a branch of two comps would be (1, 3). The `node_order` only deals with off-diagonals (as diagonal elements are handled trivially).

tests/jaxley_identical/test_basic_modules.py
michaeldeistler: I think that this is a very common use case, and one that we have been largely neglecting in our tests. So I am opting for the big test here.

jaxley/utils/solver_utils.py
jnsbck: ah, makes sense!

tests/jaxley_identical/test_basic_modules.py
jnsbck: id still argue it could be tested as part of test_net, but if you really want to separate it, feel free to go ahead with it :)

michaeldeistler: yes, I see your point here. However, this exact scenario has caused issues in the past, and I really think it is a prime use case. The tests take ~4 sec, so I think it is worth it. I might also extend this test in the future to make it more biologically relevant.

PR #703, opened 2025-08-22
Carry over groups from constituents of a module

michaeldeistler:
Fixup for things like:
```python
cell1 = jx.Cell()
cell1.add_to_group("soma")

cell2 = jx.Cell()
cell2.add_to_group("axon")

net = jx.Network([cell1, cell2])
net.soma
```

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -840,6 +840,12 @@ class Module(ABC):
             name = channel._name
             self.base.nodes.loc[self.nodes[name].isna(), name] = False
 
+        # Set columns of groups to `False` instead of `NaN`.
+        for name in self.base.group_names:
+            self.base.nodes.loc[self.nodes[name].isna(), name] = False
+            # Ensure that type is boolean---in some cases, it had been an `object`.
+            self.base.nodes[name] = self.base.nodes[name].astype(bool)
+
     @only_allow_module
     def to_jax(self):
         # TODO FROM #447: Make this work for View?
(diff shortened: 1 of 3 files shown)


jaxley/modules/base.py
```diff
@@ -840,6 +840,12 @@ def _gather_channels_from_constituents(self, constituents: List):
             name = channel._name
             self.base.nodes.loc[self.nodes[name].isna(), name] = False
 
+        # Set columns of groups to `False` instead of `NaN`.
+        for name in self.base.group_names:
+            self.base.nodes.loc[self.nodes[name].isna(), name] = False
+            # Ensure that type is boolean---in some cases, it had been an `object`.
+            self.base.nodes[name] = self.base.nodes[name].astype(bool)
```
jnsbck, line 847: could we not save one for loop here?

i.e. 
```python
for name in in self.base.channels + self.base.pumps + self.base.group_names:
    self.base.nodes.loc[self.nodes[name].isna(), name] = False
    # Ensure that type is boolean---in some cases, it had been an `object`.
    self.base.nodes[name] = self.base.nodes[name].astype(bool)
```

jnsbck (approved):
Feel free to merge this PR in any case, but if you agree with in https://github.com/jaxleyverse/jaxley/pull/703#issuecomment-3214132667 it should be easy to implement. Otherwise feel free to just open an issue and merge this for now.

...and have a look at the comment below before you do.

jaxley/modules/base.py
michaeldeistler: no, because channels need
```python
for channel in in self.base.channels
    name = channel._name
```

Issue #709, opened 2025-09-12
`get_parameters` doesn't return consistent values after updating parameters


Kartik-Sama:
I am updating initial values of trainables, if I check the values in the dataframe view values change but get_parameters method still returns old values. 
(PS: Asking because I use get_parameters to fetch parameters that are sent into optax transform, but my initial values don't seem to be updated as expected.)

```
>>> branch.nodes['radius'] 
0    1.0
1    1.0
Name: radius, dtype: float64
>>> branch.get_parameters()
[{'radius': Array([1., 1.], dtype=float32)}]
>>> branch.write_trainables([{"radius":jnp.array([0.25, 1.23], jnp.float32)}])
>>> branch.nodes['radius']
0    0.25
1    1.23
Name: radius, dtype: float64
>>> branch.get_parameters()
[{'radius': Array([1., 1.], dtype=float32)}]
```

PR #710, opened 2025-09-12
Updating the trainable_params variable  when write_trainables method

Kartik-Sama:
When the `write_trainables` function is invoked, the parameters of the module are being changed in the lines in the [`write_trainables`](https://github.com/jaxleyverse/jaxley/blob/d84b3221c1dc7f8438ad237deab680210b2f970c/jaxley/modules/base.py#L1518C9-L1522C79) method.

However when the method `get_parameters` is called (which is supposed to return only the trainable parameters) the class member [`self.trainable_params`](https://github.com/jaxleyverse/jaxley/blob/d84b3221c1dc7f8438ad237deab680210b2f970c/jaxley/modules/base.py#L188C9-L188C59) is being accessed. Which is populated in the [`make_trainable`](https://github.com/jaxleyverse/jaxley/blob/d84b3221c1dc7f8438ad237deab680210b2f970c/jaxley/modules/base.py#L1475) method, but is never updated in the flow of the `write_trainables` method. 

The code changes address [Issue 709](https://github.com/jaxleyverse/jaxley/issues/709)  - it updates the stale values being returned by `get_parameters` method when changes to trainable parameters are made through the `write_trainables` method. This done by maintaining the correct state of trainable parameters in the class member `self.trainable_params`.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1802,19 +1802,17 @@ class Module(ABC):
             v = jx.integrate(cell, params=params, t_max=10.0)
         """
         trainable_params = []
-        for inds, key in zip(
+        for comp_inds, param in zip(
             self.indices_set_by_trainables, self.trainable_param_names
         ):
-            first_inds = inds[:, 0] if inds.ndim >= 2 else inds
-            # Determine if this is a node or edge parameter
-            if key in self.nodes.columns:
-                # Fetch current values from nodes DataFrame
-                values = jnp.array(self.nodes.loc[first_inds, key])
-            elif key in self.edges.columns:
-                values = jnp.array(self.edges.loc[first_inds, key])
-            else:
-                raise KeyError(f"Parameter '{key}' not found in nodes or edges.")
-            trainable_params.append({key: values})
+            data = self.nodes if param in self.nodes.columns else None
+            data = self.edges if param in self.edges.columns else data
+            assert data is not None, f"Parameter '{param}' not found in nodes or edges"
+            param_vals = jnp.nanmean(
+                jnp.asarray([data.loc[inds, param].to_numpy() for inds in comp_inds]),
+                axis = 1
+            )
+            trainable_params.append({param: param_vals})
         return trainable_params
 
     @only_allow_module


kyralianaka:
Hello there, yes I initialize trainable parameters using make_trainable() currently everywhere, and I think it's a major API change that could at least go to the v1 release. Maybe there is a way to still delete ```self.trainable_params``` but rather call ```set()``` if initial values are provided?

michaeldeistler:
Okay, then let us not remove the `init_value` option.

However, maybe the easiest option is the one that @Kartik-Sama had originally implemented: to just write to `self.trainable_params` in `write_trainables()`. @Kartik-Sama do you see any major downsides to this?

Kartik-Sama:
I see that trainable parameters values are being stored in two places 1) dataframe corresponding to the module's base 2) `self.trainable_params`

Having both of them is fine as long as both these places are consistently updated whenever changes happen. But the cleaner way would be to just have one place to store this information and retrieve it in the format necessary. We can still get rid of `trainable_params` member while having the `init` in `make_trainable` though.  Directly updating the code in `write_trainable` is an easy fix, but still looks like a shaky fix to me.

I also have doubts with a [test](https://github.com/jaxleyverse/jaxley/blob/a368f5be6efa801cbf1c8ad45710e97f40582ec1/tests/test_make_trainable.py#L85C5-L85C31) though. Here the params are being fetched from get_params twice (at lines [98](https://github.com/jaxleyverse/jaxley/blob/a368f5be6efa801cbf1c8ad45710e97f40582ec1/tests/test_make_trainable.py#L98) and [120](https://github.com/jaxleyverse/jaxley/blob/a368f5be6efa801cbf1c8ad45710e97f40582ec1/tests/test_make_trainable.py#L120)). However in between these lines I see no writing or setting of `trainable_params` variable, yet this [assertion](https://github.com/jaxleyverse/jaxley/blob/a368f5be6efa801cbf1c8ad45710e97f40582ec1/tests/test_make_trainable.py#L127) checks for the new value of the parameter without being actually set. I am definitely missing something here, but the test itself isn't very clear to me.

Kartik-Sama:
`trainable_params` can be split into two chunks of information. 
1) The names of trainable parameters
2) Indices where you have to look up the parameter values (which is already being stored in indices_set_by_trainables)

In that way we can always call `get_parameters` to load the current state of trainable parameters, since we know the column and the indices (and thereby using .loc) to retrieve or set their values.

For instance the `write_trainables` works without ever setting the `trainable_params` because it is updating the underlying data frame correctly. Thus, we can just convert the `init_val` to appropriate format and just call `write_trainables` method from inside `make_trainable` method. (Unless I am missing something)

PR #711, opened 2025-09-16
Bugfix for indexing in `init_states` for networks

michaeldeistler:
To reproduce the error:

```python
from jaxley.channels import HH

comp = jx.Compartment()
branch = jx.Branch(comp, ncomp=2)
cell1 = jx.Cell(branch, [-1, 0, 1])
cell2 = jx.Cell(branch, [-1, 0, 1, 2])
cell3 = jx.Cell(branch, [-1, 0])
net = jx.Network([cell1, cell2, cell3])

net.cell(0).branch(1).add_to_group("apical")
net.cell(1).branch(3).add_to_group("apical")
net.cell(2).branch(1).add_to_group("apical")

net.apical.insert(HH())

net.init_states()
```

The reason that this bug happened is that `init_states` was still using the `global_comp_index`, instead of the `index` (as is done in `step_channels`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # 0.11.4 (pre-release)
 
+### 🐛 Bug fixes
+
+- bugfix for indexing when `init_states()` is run on a `jx.Network` (#711,
+@michaeldeistler)
+
 ### 📚 Documentation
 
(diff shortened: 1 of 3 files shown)


jnsbck (approved):
Looks good! Thanks for adding a test. Good to go.

PR #713, opened 2025-09-16
Fixup for `.delete()` when multiple channels have the same current_name

michaeldeistler:
Currently, `.delete()` deletes the `current_name` when a channel is deleted, even if another channel also has that same `current_name`. Same for shared parameters or shared states.

This PR fixes this.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -2246,7 +2246,39 @@ class Module(ABC):
         """Remove a channel or pump from the module.
 
         Args:
-            channel: The channel to remove."""
+            channel: The channel to remove.
+
+        Example usage
+        ^^^^^^^^^^^^^
+
+        The example below inserts two channels and then deletes the sodium channel:
+
+        ::
+
+            from jaxley.channels import Na, K
+
+            cell = jx.Cell()
+            cell.insert(Na())
+            cell.insert(K())
+
+            cell.delete(Na())
+
+        The example below inserts two channels and then deletes both of them:
+
+        ::
+
+            from jaxley.channels import Na, K
+
+            cell = jx.Cell()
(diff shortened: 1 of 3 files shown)


jaxley/modules/base.py
```diff
@@ -2270,8 +2302,13 @@ def delete(self, channel: Union[Channel, Pump]):
                         "`cell.delete(HH())` (ie you forgot to initialize the channel "
                         "via round brackets: `HH()`."
                     )
-                self.base.membrane_current_names.remove(channel.current_name)
+
                 self.base.nodes.drop(columns=channel_cols + [name], inplace=True)
+
+            # Re-compute the current names.
+            self.base.membrane_current_names = list(
```
jnsbck, line 2320: Good catch. Could this happen for shared params as well? I.e. Would the shared_gNa column be removed entirely below or not?

```python
na1 = Na("Na1")
na1.channel_params["shared_gNa"] = na1.channel_params.pop("Na1_gNa")
na2 = Na("Na2")
na2.channel_params["shared_gNa"] = na1.channel_params.pop("Na2_gNa")

cell = jx.Cell()
cell.insert(na1)
cell.insert(na2)

cell.delete(na1) # what happens ?
```

If yes, we might wanna harden against this as well?

jnsbck: I think the check in L2293 might not catch this case.

jnsbck (changes requested):
I have a small concern, that would be great to have checked before this is merged. Otherwise good catch and fine to merge.

jaxley/modules/base.py
michaeldeistler: Great catch, fixed it now!

PR #714, opened 2025-09-18
safe_softplus

Matthijspals:
Our implementation of `softplus` in parameter transforms makes it return 20.0 if the parameter is larger than 20.0, as we implement it as `jnp.log1p(save_exp(x))` and `save_exp` clips values above 20.0. 

A safe way to counter this is to just use the linear function if the input is bigger than a certain threshold (e.g., 20.0 like the save_exp). This is also what is done in PyTorch for example.

For example:

```
def softplus_save(x, lower=0.0):
    return jnp.where(x > 20.0, x, jnp.log1p(save_exp(x))) + lower

def softplus(x, lower=0.0):
    return jnp.log1p(save_exp(x)) + lower

print(softplus_save(25))
print(softplus(25))
```
results in
```
25.0
20.0
```
25.0 is desirable I'd think!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,9 +20,9 @@
 
 # -- Project information -----------------------------------------------------
 
-project = "Jaxley"
-copyright = "2024, Jaxleyverse team"
-author = "Jaxleyverse team"
+project = 'Jaxley'
+copyright = '2024, Jaxleyverse team'
+author = 'Jaxleyverse team'
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,23 +1,31 @@
...
+### ??? Bug fixes
+
+- safe softplus, use linear function above certain threshold. This avoids an unwanted clipping 
+operation due to the save_exp (#714 @matthijspals)
+
+
 # 0.11.4
 
-### 🐛 Bug fixes
+### ???? Bug fixes
```
michaeldeistler: Something happened here :D

Matthijspals: ah I think it was because I edited it with nano which didn't like the smileys

michaeldeistler:
Looks great now, thanks for the fix!

PR #727, opened 2025-10-16
separate currents from `get_all_states()`

michaeldeistler:
In discussion with @Matthijspals, we found that this PR makes #719 easier to implement. I also think that separating this logic is nice and clearer.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -1695,7 +1695,10 @@ class Module(ABC):
         # The value for `delta_t` does not matter here because it is only used to
         # compute the initial current. However, the initial current cannot be made
         # trainable and so its value never gets used below.
-        all_states = self.base.get_all_states(pstate, all_params, delta_t=0.025)
+        all_states = self.base.get_all_states(pstate)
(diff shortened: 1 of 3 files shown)


jaxley/modules/base.py
```diff
@@ -1926,19 +1929,19 @@ def _get_states_from_nodes_and_edges(self) -> dict[str, Array]:
         return states
 
     @only_allow_module
-    def get_all_states(
-        self, pstate: list[dict], all_params, delta_t: float
-    ) -> dict[str, Array]:
+    def get_all_states(self, pstate: list[dict]) -> dict[str, Array]:
```
jnsbck, line 1932: on 2nd thought we could add a warning here / that the behaviour changed for a few releases.

jnsbck: might be annoying though since I just noticed it would be triggered on every integrate...

michaeldeistler: I think `get_all_states` is really barely user-facing (not mentioned in any tutorial, how-to guide, etc), so I think it is fine to just mention this in the CHANGELOG.md.

PR #730, opened 2025-10-23
feat: Utilities to extract the dynamic (ODE) states of a module

### LINKED ISSUE #717: Dynamics function
It would be great to have a dynamics function `F` that allows one to iteratively step through Jaxley networks like `x_{t+1}=F(x_t, u_t)`, where `x` is a vector of _states_ (excluding _observables_ like currents), and `u` is a vector of input.

 This has many use-cases, such as filtering / inference algorithms (see Discussion #715), calculating Jacobians (see Discussion #699), stability analysis (e.g., our paper figure where we plot Lyaponuv exponents) and many more.

Discussion #715 contains some code by both @bantin and me, that aims to achieve this, and I can soon open a pull-request based on that. 

Some things to consider:
**1.** Where do we want to have this code:
Quoting @michaeldeistler:
> - We could have them as a small module as part of Jaxley/utils/
> - We could open a new repo called Jaxely-utilities that implements such things?
> - Or we simply write a clean example on how to do this, and expect people to copy relevant code pieces.
> To get started, I tend towards the first option, at least as long as it is rather little code. As the code base grows bigger, we might want to migrate to option 2.

I agree with this statement and think it makes sense to add it to Jaxley for now. Maybe in the future, we can have a separate repo with utilities that include e.g., inference algorithms.

**2.** Implementation
The reason we can not use the current `step_fn`, is that, it is not a vector but a dict (easy to manage with `ravel_pytree`), and:
 - 2.1 It contains values that are not true _states_, but _observables_ in the dynamical system sense, such as the currents.
 - 2.2 It contains branch-points, which if I understand correctly can be recomputed based on the states of the non-branch points (please confirm or correct!)
 - 2.3 It is padded with NaNs if a state is only relevant for some compartments.

2.1 Can be straightforwardly addressed by splitting the `step_fn` in the part that computes the states, and the part that computes the currents (see e.g., my code in Discussion #715)
2.3 should also not be too hard, we can simply store the indices of the NaN elements, remove them and add them back where needed (maybe there is some useful` jax pytree` functionality for this).
2.2 not 100% sure about branch points, but I guess we similarly filter them out, and then recompute them when needed.

michaeldeistler:
Closes #717

Code in this PR was by and large written by @Matthijspals.

This PR adds utilities to process the ``all_states`` dictionary such that it only contains "true" ODE states:
- it removes branchpoint states
- it removes membrane states that are NaN because the channel does not exist there
- it removes ionic currents.

In addition, the utilities can flatten the states to a vector.

These utilities make it easier to build, for example, the jacobian, and they also make interoperability with frameworks like [dynamax](https://probml.github.io/dynamax/) easier.

At its core, this functionality is executed by four utilities:
```python
from jaxley.utils.dynamics import build_dynamic_state_utils
remove_observables, add_observables, flatten, unflatten = build_dynamic_state_utils(cell)
```

Simplest usage is:
```python
import jaxley as jx
from jaxley.integrate import build_init_and_step_fn
from jaxley.utils.dynamics import build_dynamic_state_utils

cell = jx.Cell()

init_fn, step_fn = build_init_and_step_fn(cell)
remove_observables, add_observables, flatten, unflatten = build_dynamic_state_utils(cell)

all_states, all_params = init_fn(params)

dynamic_states = flatten(remove_observables(all_states))
recovered_all_states = add_observables(unflatten(dynamic_states), all_params, delta_t=0.025)
```

### DIFF
--- a/jaxley/utils/dynamics.py
+++ b/jaxley/utils/dynamics.py
@@ -44,13 +44,13 @@ def build_dynamic_state_utils(module) -> Tuple[Callable, Callable, Callable, Cal
 
     .. warning::
 
-    If a membrane current is used as a state by another channel (for example,
-    the calcium current ``i_ca`` used in the ``channel_states`` of a
-    calcium-dependent potassium channel), then this current will be included
-    in the returned "true" ODE states. Similarly, if there are ``channel_states``
-    that are directly computed based on other states (e.g., the calcium reversal
-    potential via the nernst equation), they also be considered "true" ODE states. We
-    recommend removing such currents manually.
+        If a membrane current is used as a state by another channel (for example,
+        the calcium current ``i_ca`` used in the ``channel_states`` of a
+        calcium-dependent potassium channel), then this current will be included
+        in the returned "true" ODE states. Similarly, if there are ``channel_states``
+        that are directly computed based on other states (e.g., the calcium reversal
+        potential via the nernst equation), they are also considered "true" ODE states.
+        Please remove such states manually if desired.
 
     Args:
         module: A ``Module`` object, e.g., a ``jx.Cell``.


jaxley/utils/dynamics.py
```diff
@@ -0,0 +1,346 @@
...
+
+    This can be used used to perform step-by-step updates where states are vector
+    representations of the full state pytree of the module, and only contain
+    true dynamical states (e.g. voltages and gating variables), and no observables
+    (e.g. currents). This is useful for e.g. computing Jacobians and Kalman filters.
+
+    Args:
+        module: A `Module` object that e.g. a cell.
+        params: Trainable parameters of the neuron model.
+        param_state: Parameters returned by `data_set`.. Defaults to None.
```
michaeldeistler: `..` typo

jaxley/modules/network.py
```diff
@@ -336,6 +336,10 @@ def _synapse_currents(
     ) -> tuple[dict, tuple[Array, Array]]:
         voltages = states["v"]
 
+        current_states = {}
```
jnsbck: potentially a dumb question, but whats `current_states`?

jnsbck: its unused?

jaxley/utils/dynamics.py
```diff
@@ -0,0 +1,537 @@
...
+
+          * Args:
+
+            * The dynamic states of the system as a flat jax array (Array).
+
+          * Returns:
+
+            * Dynamic states as a dict of Arrays (Dict[str, Array]).
+
+    Example usage
```
jnsbck, line 114: thats quite extensive, should we add a how to? to the docs?

jaxley/utils/dynamics.py
```diff
@@ -0,0 +1,537 @@
...
+        if _is_valid_membrane_leaf(key, leaf1, valid_keys):
+            return fn(leaf1, leaf2)
+        else:
+            # Pass through leaf1 unchanged otherwise.
+            return leaf1
+
+    return tree_map_with_path(wrapper, tree1, tree2)
+
+
+def _get_key_name(path):
```
jnsbck, line 519: Does this assume that no 2 leaves can be the same, or would something like `Na().channel_params{"gbar": 1.0}` and  `Na().channel_params{"gbar": 1.0}` potentially cause issues with all the tree_mapping?

jaxley/modules/network.py
Matthijspals: Good question, the changes to `network.py `were added in the `proposed new api; not tested ` commit. I can't remember adding this (although I might have forgotten), did you remember if you added this @michaeldeistler? 

I think we could just remove all changes to network.py.

jaxley/utils/dynamics.py
Matthijspals: What exactly do you mean with`Na().channel_params{"gbar": 1.0}` and  `Na().channel_params{"gbar": 1.0}` ? One compartment with two Na channels?

jaxley/modules/network.py
Matthijspals: I quickly checked locally, and it seems like indeed all changes to `network.py` can be reverted, but let's wait for Michael to confirm.

Matthijspals: I am also not sure about the change to `base.py`

jaxley/utils/dynamics.py
jnsbck: Ah sorry copy paste error. It was sposed to be `Na().channel_params{"gbar": 1.0}` and  `Na2().channel_params{"gbar": 1.0}`, i.e. two channels with the same param name

jaxley/modules/network.py
michaeldeistler: Sorry, I had added this, but it is not used. I will delete it.

We do, however, need the later bits:
```python
current_name = f"i_{synapse_type._name}"
if current_name not in self.base.synapse_current_names:
    self.base.synapse_current_names.append(current_name)
```
These update `synapse_current_names`, which is used in the `dynamics.py` to delete the currents through synapses from the `all_states` dictionary.

jaxley/utils/dynamics.py
michaeldeistler: My take is that the API reference should be the only part of the docs where we aim for **completeness**. I think these dynamics utils are nice, but probably don't yet warrant a how-to guide.

jaxley/utils/dynamics.py
michaeldeistler: I am confused here as well. Do you mean a "global" channel parameter (like reversal potential)? If parameters have the same name then they automatically get considered as the same param.

jnsbck: yea! ok, just checking that this is not an edge case that messes with things.

Matthijspals: @jnsbck Could you maybe send the lines of code that would construct a cell with this edge case? Just want to make sure I understand you properly.

michaeldeistler: I have done some testing. I found a breaking case when `currents` are _also_ used as states in other channels, but this is fixed, tested, and documented now.

PR #743, opened 2025-11-28
Exponential Euler and forward Euler solvers

michaeldeistler:
This PR implements an exponential Euler solver, and it implements forward Euler also for branched morphologies. Thanks to Amrith Lothikar for discussions!

### API

Most simple API for running exponential Euler
```python
jx.integrate(cell, solver="exp_euler")
```

API for pre-computing the diffusion terms (and then not computing them during `jx.integrate()`):
```python
cell.customize_solver_exp_euler(
    exp_euler_transition=cell.build_exp_euler_transition_matrix(delta_t)
)
jx.integrate(cell, solver="exp_euler")
```
This can be useful if the diffusion terms (radius, length, ra, C) are the same across many simulations.

### Major limitations

The biggest limitation is that the solver does not yet understand the concept of independent neurons. That is, it treats networks as one large cell and performs matrix-vector multiplications, instead of block-diagonal matrix-vector multiplications (or even vector-vector for networks of single compartment).

### Benchmark

This solver requires to solve a matrix exponential upfront (only once for a given `axial_resistivity`, `length`, `radius`), and then requires _one_ matrix vector product at every integration step. Thus:
- this solver can be very performant on GPU.
  - If the matrix exponential can be pre-computed:
    - For single cells and `batchsize=1`, I get ~10x speedups. 
    - For `batchsize=10_000`, I get ~2x speedups.
  - If the matrix exponential cannot be pre-computed, but it is the same for every element in the entire batch (e.g., because the batch vectorizes across stimuli). This case requires additional effort from the user (see below):
    - The additional cost of pre-computing the matrix is small.
  - If the matrix exponential cannot be pre-computed and it is different for every element in the batch:
    - For small morphologies (~500 compartments), the matrix exponential takes as long as ~1 second of simulation.
    - For large morphologies (>1000 compartments), the cost of the matrix exponential becomes excessive and the solver becomes essentially unusable.
- the solver is very slow on CPU (around 10x slow-down).

### Summary

Another benefit is that this solver produces more accurate solutions than backward Euler, especially for very large `delta_t`. Even at `dt=1.0 ms`, I get somewhat good looking voltage traces and spikes.

Overall, I think this PR will allow adding some scale by making GPU runs faster (especially if the transition matrix can be precomputed), and potentially allowing a larger `delta_t`, at least for initial explorations.

### Appendix: How to implement the same exp euler matrix for every batch element

```python
def simulate(amp, euler_matrix):
    cell.customize_solver_exp_euler(exp_euler_transition=euler_matrix)
    dstate = cell.branch(0).comp(0).data_stimulate(jx.step_current(100.0, 840.0, amp, delta_t, t_max))
    return jx.integrate(cell, data_stimuli=dstate, solver="exp_euler")

vmapped_simulate = vmap(simulate, in_axes=(0, None))

def loss(ra, amps):
    pstate = cell.data_set("axial_resistivity", ra, None)
    cell.to_jax()
    all_params = cell.get_all_parameters(pstate)
    axial_conductances = all_params["axial_conductances"]["v"]
    euler_matrix = cell.build_exp_euler_transition_matrix(
        delta_t=delta_t, axial_conductances=axial_conductances
    )
    return vmapped_simulate(amps, euler_matrix)

jitted_loss = jit(loss)

start_time = time.time()
ra = jnp.asarray(np.random.rand() * 1000)
amps = jnp.asarray(np.random.rand(num_vmap))
v = jitted_loss(ra, amps)
v.block_until_ready()
logging.info(f"EXP Compile Time: {time.time() - start_time}")
```

### DIFF
--- a/jaxley/integrate.py
+++ b/jaxley/integrate.py
@@ -255,12 +255,8 @@ def integrate(
             with zeros. If `t_max` is smaller, then the stimulus with be truncated.
         delta_t: Time step of the solver in milliseconds.
         solver: Which ODE solver to use. Either of ["fwd_euler", "bwd_euler",
-            "crank_nicolson", "exp_euler"]. The solvers can be customized before
-            running `jx.integrate` with the `module.customize_solver_...` methods.
-            For example: `cell.customize_solver_bwd_euler(voltage_solver='jaxley.dhs')`.
-            Note that `exp_euler` is still experimental: It does not work for
-            morphologies that have zero channels, and it is not yet optimized for
-            networks.
+            "crank_nicolson", "exp_euler"]. Note that `exp_euler` is still
+            experimental and is not yet optimized for networks.
         voltage_solver: Algorithm to solve quasi-tridiagonal linear system describing
             the voltage equations. The different options only take effect when
             `solver` is either `bwd_euler` or `crank_nicolson`. The options for
@@ -311,6 +307,7 @@ def integrate(
 
         cell._init_morph_jaxley_dhs_solve(allowed_nodes_per_level=16)
         v = jx.integrate(cell, voltage_solver="jaxley.dhs.gpu")
+
     """
     if voltage_solver == "jaxley.dhs":
         # Automatically infer the voltage solver.


docs/tutorials/04_jit_and_vmap.ipynb
```diff
@@ -416,6 +416,34 @@
...
+    "cell.customize_solver_exp_euler(\n",
+    "    exp_euler_transition=cell.build_exp_euler_transition_matrix(delta_t)\n",
+    ")\n",
+    "\n",
+    "v1 = jx.integrate(cell, solver=\"exp_euler\")\n",
+    "cell.set(\"gLeak\", 1e-4)\n",
+    "v2 = jx.integrate(cell, solver=\"exp_euler\")\n",
+    "```\n",
+    "\n",
+    "With these modifications, the exponential Euler can sometimes by 10 times faster than backward Euler **on GPU**, so it can be worth exploring this."
```
jnsbck: typo, "be"

jaxley/integrate.py
```diff
@@ -366,6 +371,15 @@ def integrate(
     )
     all_states, all_params = init_fn(params, all_states, param_state, delta_t)
 
+    if solver == "exp_euler":
```
jnsbck, line 371: if this is still experimental, maybe we should emit a warning and ask users to report bugs.

jaxley/modules/base.py
```diff
@@ -1911,6 +1918,266 @@ def get_all_parameters(self, pstate: list[dict]) -> dict[str, Array]:
...
+        .. code-block:: python
+
+            delta_t = 0.025
+            cell = jx.Cell()
+            cell.customize_solver_exp_euler(
+                exp_euler_transition=cell.build_exp_euler_transition_matrix(delta_t)
+            )
+            v = jx.integrate(cell, delta_t=delta_t, t_max=100.0)
+        """
+        if exp_euler_transition is not None:
```
jnsbck, line 2049: should we check if `r` etc. is in trainables and warn the user?

jaxley/modules/base.py
```diff
@@ -1911,6 +1918,266 @@ def get_all_parameters(self, pstate: list[dict]) -> dict[str, Array]:
...
+            )
+        else:
+            g_branchpoint = jnp.asarray([])
+        # Step 4: Concatenate diagonals, comp-to-comps, and branchpoint influences.
+        vals = jnp.concatenate([g_diags, g_offdiags, g_branchpoint])
+        rows = indexer["rows"]
+        cols = indexer["cols"]
+        return vals, rows, cols
+
+    def build_exp_euler_transition_matrix(
```
jnsbck, line 2123: you could add `@only_allow_module` just to be safe

jaxley/solver_voltage.py
jnsbck, line 1: Since we now support so many solvers, should we also test them against each other, i.e. take bwd_euler with very small steps as ground truth and force deviation of others to be small?

jaxley/integrate.py
jnsbck, line 371: didnt we have a device variable somewhere? Could be useful info to check if user is running on CPU and warn them that this might be suboptimal if they use `exp_euler`.

docs/tutorials/04_jit_and_vmap.ipynb
jnsbck, line 1: Should we add a "how to choose a solver (and stepsize)" on the how to section?

I think there is a lot of useful info contained in the PR description for example that would otherwise be lost.

jaxley/integrate.py
michaeldeistler: I would prefer to not have a warning (they are just annoying...). Since it is not default I think it is fine.

jaxley/modules/base.py
michaeldeistler: I don't think we should, because it's very hard to do this check: A user might have done `.set()`, `.data_set`, `.make_trainable` (or even hard coded changes into the .nodes). Since this `if-case` is not something that gets run by default, I am not sure if it is worth adding all of this additional logic to cover all of these cases.

jaxley/integrate.py
michaeldeistler: Hm, not sure...it depends on so many things, so I am not sure we should really give this advice. And I just don't like having tons of warning everywhere :)

docs/tutorials/04_jit_and_vmap.ipynb
michaeldeistler: I think this could be useful, but I feel like we should still get a bit more feedback before writing such a how-to guide (and run more tests internally).

michaeldeistler: i ended up writing a new how-to-guide

PR #750, opened 2025-12-12
feat: verbose recordings

### LINKED ISSUE #554: named recordings
```python
cell.record()
v = jx.integrate(cell)  # will return N voltages, but which one corresponds to which one?
```

Solution:
```python
from jx.utils import prettify
v = jx.integrate(cell)
v = prettify(v, cell)

# -> v = {"v_cell0_branch0_comp0": ..., "v_cell0_branch0_comp1": ....}
```

michaeldeistler:
Closes #554 

```python
import jaxley as jx

net = jx.Network(...)

net.cell(0).branch(0).comp(0).record()
net.IonotropicSynapse.edge(0).record("IonotropicSynapse_s")
recs = jx.integrate(net, t_max=10.0)

net.write_recordings(recs)
print(net.cell(0).branch(0).comp(0).recording())  # Shape (1, T).
print(net.IonotropicSynapse.edge(0).recording("IonotropicSynapse_s"))  # Shape (1, T).
```

### DIFF
--- a/docs/jaxley.rst
+++ b/docs/jaxley.rst
@@ -137,7 +137,6 @@ Utilities
    :toctree: reference/_autosummary
    :nosignatures:
 
-   jaxley.utils.verbose_recordings
    jaxley.solver_gate.save_exp
    jaxley.solver_gate.solve_gate_implicit
    jaxley.solver_gate.solve_gate_exponential


michaeldeistler:
EDIT: Maybe it would be better to have a `write_recordings_to_nodes`?

jaxley/utils/api_utils.py
```diff
@@ -0,0 +1,48 @@
...
+        cell.record("v", prefix="excitatory_")
+        v = jx.integrate(cell, t_max=10.0)
+
+        recs = verbose_recordings(v, cell.recordings)
+        print(recs["excitatory_v_0"])  # Array of shape (T,)
+
+    """
+    verbose_recs = {}
+    for i, rec in module_recordings.iterrows():
+        verbose_recs[f"{rec['prefix']}{rec['state']}_{rec['rec_index']}"] = recordings[
```
jnsbck: underscore after prefix?

jaxley/utils/api_utils.py
```diff
@@ -0,0 +1,48 @@
+# This file is part of Jaxley, a differentiable neuroscience simulator. Jaxley is
+# licensed under the Apache License Version 2.0, see <https://www.apache.org/licenses/>
+
+import pandas as pd
+from jax import Array
+
+
+def verbose_recordings(
```
jnsbck: Could we not just make recordings a property of base and then return it as a dict?

I would have suggested sth like:

```python
@property
def recordings(self):
    recs = {f"{prefix}_{state}_{idx}": rec for idx, state, prefix, rec in zip(self.recs.iterrows(), self.recordings)}
    return recs
```

jaxley/utils/api_utils.py
jnsbck: otherwise the user has to think about adding it, that could be annoying

michaeldeistler: I disagree---if we manually add the prefix then there is no way of getting rid of it anymore for the user. In particular, without any prefix there will also be a _ ahead of `state`. We can of course add logic to deal with this, but I don't think it's worth it (not even pandas does it for you :) )

jnsbck: But filtering will be a pain otherwise, with  `f"{rec['prefix']}{rec['state']}_{rec['rec_index']}"` i.e. for `prefix="gate"` and the  `state="s"` this would yield `"states_0"` rather than `"state_s_0"`. In the latter case one could just `.split("_")`, which I think is more consistent.

michaeldeistler: yes, but users can also just do `prefix = 'gate_'`. As I said above, pandas is also doing it this way, so I would suggest we stay consistent. But either way, the `prefix` option is removed in the most recent version of this PR :)

PR #753, opened 2025-12-12
feat: external states

michaeldeistler:
This PR provides utilities to provide external states. I can imagine two use-cases:
- time varying channel conductances: can be implemented by doing:
```
cell.external_state("multiplier", jnp.linspace(0, 100, 1000))
```
and then using `gbar * states["multiplier"]` within the `Channel`.
- testing receptors without having to instantiate pre-synaptic neurons. E.g., we can let the `external_state` be a poisson spike train and then have a synapse that convolves this with some dynamics.

However, I am not yet dead-convinced that we need this. One could also work around these issues in a different way.

### DIFF
--- a/jaxley/modules/base.py
+++ b/jaxley/modules/base.py
@@ -203,6 +203,8 @@ class Module(ABC):
         # E.g. `self.external)inds = {"v": jnp.asarray([0,1]), "i": jnp.asarray([2,3])}`
         self.external_inds: dict[str, ArrayLike] = {}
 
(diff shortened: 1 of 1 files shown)


WillemWybo:
Ah interesting, I assumed a clamp would work like in neuron by injecting a clamping current. Does a clamp fully replace the dynamics? How does this work if there are coupled compartments in the cell (sorry, this discussion is probably a bit off topic here)

PR #754, opened 2025-12-12
Init params

### LINKED ISSUE #507: `def init_params()` method for channels
In many cases, one would want parameters to depend on one another. For example, if one has
```python
channel_params = {"temperature": 37, "gNa": 0.1}
```
then the sodium conductance would depend on temperature via:
```python
effective_gNa = 2.3**(temperature - 37) * gNa
```

I think that the most elegant way of making this possible is to add a `init_params()` method to the channel.
```python
def init_params(self, states, v, params):
    effective_gNa = 2.3**(params["temperature"] - 37) * params["gNa"]
    return {"gNa": effective_gNa}
```

michaeldeistler:
Closes #507 

This PR adds an `init_params` method to channels.
```python
cell = jx.Cell()
cell.insert(HH())
cell.set("celsius", 27.0)  # Reduce temperature by 10 degrees.
cell.init_params()
cell.nodes
```

### DIFF
--- a/jaxley/channels/hh.py
+++ b/jaxley/channels/hh.py
@@ -26,7 +26,7 @@ class HH(Channel):
             f"{prefix}_eK": -77.0,
             f"{prefix}_eLeak": -54.3,
             f"{prefix}_tadj": 1.0,
-            f"celsius": 37.0,
+            f"temperature": 37.0,  # Celsius.
         }
         self.channel_states = {
             f"{prefix}_m": 0.2,
@@ -94,7 +94,7 @@ class HH(Channel):
         """Initialize the maximal conductances given the temperature."""
         prefix = self._name
         q10 = 2.3
-        t = params["celsius"]
+        t = params["temperature"]
         tadj = q10 ** ((t - 37.0) / 10.0)
         return {f"{prefix}_tadj": tadj}
 
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -592,6 +592,6 @@ def test_multicompartment_izhikevich(SimpleCell):
 def test_init_params():
     cell = jx.Cell()
     cell.insert(HH())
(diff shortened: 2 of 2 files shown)


jaxley/channels/channel.py
```diff
@@ -109,3 +109,12 @@ def init_state(
     ):
         """Initialize states of channel."""
         return {}
+
+    def init_params(
```
jnsbck, line 113: should we document this in the [channel tutorial](https://github.com/jaxleyverse/jaxley/blob/main/docs/tutorials/05_channel_and_synapse_models.ipynb) ?

jaxley/channels/hh.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, name: Optional[str] = None):
             f"{prefix}_eNa": 50.0,
             f"{prefix}_eK": -77.0,
             f"{prefix}_eLeak": -54.3,
+            f"{prefix}_tadj": 1.0,
+            f"celsius": 37.0,
```
jnsbck: how about `"temperature": 37.0 # celsius` ? I dont like that celsius is rather a unit

jaxley/channels/hh.py
```diff
@@ -83,6 +85,19 @@ def init_state(
             f"{prefix}_n": alpha_n / (alpha_n + beta_n),
         }
 
+    def init_params(
```
jnsbck, line 88: How do you feel about something like `init` and `set`?  This would also mimic what we have in module. `set` can manually change any state or param, while `init` does so based on predefined rules.

jnsbck: I am also happy with `init_states`, `init_params` though. Should we also have an `update_params` btw?

michaeldeistler: I would argue that a param that can be updated should rather be a state.

PR #765, opened 2025-12-21
New synapse API

michaeldeistler:
```python
class CurrentSynapse(Synapse):
    def __init__(self, name: Optional[str] = None):
        super().__init__(name)
        prefix = self._name
        self.synapse_params = {}
        self.synapse_states = {}
        self.node_params = {"e_syn": 0.0}
        self.node_states = {"ca_i": 0.0}

    def update_states(
        self,
        synapse_states: dict[str, Array],
        synapse_params: dict[str, Array],
        pre_voltage: Array,
        post_voltage: Array,
        pre_states: dict[str, Array],
        post_states: dict[str, Array],
        pre_params: dict[str, Array],
        post_params: dict[str, Array],
        delta_t: float,
    )
```

I decided to also give all synapses access to presynaptic _parameters_. I think this will be super useful, because it makes parameter sharing very easy: e.g., the reversal potential should really _always_ be shared across post-synaptic cell types, and never across individual synapses. This is now super easy:
```python
net.cell([0, 1]).add_to_group("excitatory")
net.cell([2, 3]).add_to_group("inhibitory")

connect(..., DynamicSynapse())

net.excitatory.set("DynamicSynapse_e_syn", 20.0)
net.inhibitory.set("DynamicSynapse_e_syn", -70.0)
```


CHANGELOG.md
```diff
@@ -38,6 +40,22 @@ import jax.numpy as jnp
...
+self,
+synapse_states: dict[str, Array],
+synapse_params: dict[str, Array],
+pre_voltage: Array,
+post_voltage: Array,
+pre_states: dict[str, Array],
+post_states: dict[str, Array],
+pre_params: dict[str, Array],
+post_params: dict[str, Array],
+delta_t: float,
```
jnsbck, line 56: is delta_t what we are going for now or dt?

CHANGELOG.md
```diff
@@ -38,6 +40,22 @@ import jax.numpy as jnp
 from jaxley.synapses import CurrentSynapse
 connect(..., CurrentSynapse(jnp.tanh))
 ```
+- Synapses have a new API (#765, @michaeldeistler). The `update_states` method and the
+`compute_current` method should both receive all of the following arguments, in that
+order:
+```python
+self,
+synapse_states: dict[str, Array],
```
jnsbck, line 48: does it make sense to do sth similar for channels? i.e. 
```
channel_params,
channel_states,
voltage,
states,
params,
```

jnsbck: ... also does that mean we commit to [mech_states, mech_params, v, states, params, delta_t] as the order of args for channels, synapses etc?

docs/tutorials/05_channel_and_synapse_models.ipynb
```diff
@@ -311,30 +311,61 @@
     "    \"\"\"\n",
     "    def __init__(self, name = None):\n",
     "        super().__init__(name)\n",
-    "        self.synapse_params = {\"gChol\": 0.001, \"eChol\": 0.0}\n",
+    "        self.synapse_params = {\"gChol\": 0.001}\n",
     "        self.synapse_states = {\"s_chol\": 0.1}\n",
-    "\n",
-    "    def update_states(self, states, all_states, pre_index, post_index, params, delta_t):\n",
+    "        self.node_params = {\"eChol\": 0.0}\n",
+    "        self.node_states = {}\n",
```
jnsbck, line 317: I dont 100% like `node_states` tbh. I think something like `accessible_comp_params` could be more descriptive (I admit, that comp could be confusing though since channels params can also be accessed). Maybe just `accessible_params`. Or something. We could get rid of the channel filtering via the same mechanics btw. Should we?

jaxley/integrate.py
```diff
@@ -57,8 +57,8 @@ def build_init_and_step_fn(
         params = cell.get_parameters()
 
         cell.to_jax()
-        rec_inds = cell.recordings.rec_index.to_numpy()
-        rec_states = cell.recordings.state.to_numpy()
+        rec_inds = cell.rec_info.rec_index.to_numpy()
```
jnsbck, line 60: why change this?

docs/tutorials/05_channel_and_synapse_models.ipynb
jnsbck: I am torn... I get why node_states, but "nodes" are not really a concept that the user is introduced to at any point.

jaxley/modules/network.py
```diff
@@ -311,15 +311,30 @@ def _step_synapse_state(
             pre_indices = np.asarray(pre_syn_inds[synapse_names[i]])
```
jnsbck, line 311: does this work with numpy?

jaxley/modules/network.py
```diff
@@ -367,19 +382,36 @@ def _synapse_currents(
...
             # Compute slope and offset of the current through every synapse.
             pre_v_and_perturbed = jnp.stack(
                 [voltages[pre_inds], voltages[pre_inds] + diff]
             )
             post_v_and_perturbed = jnp.stack(
                 [voltages[post_inds], voltages[post_inds] + diff]
             )
-            comp_cur_fn = vmap(synapse_type.compute_current, in_axes=(0, 0, 0, 0))
-            synapse_currents = vmap(comp_cur_fn, in_axes=(None, 0, 0, None))(
+            comp_cur_fn = vmap(synapse_type.compute_current, in_axes=(0, 0, 0, 0, 0, 0, 0, 0, None))
```
jnsbck, line 404: one could define the in_axes on a seperate line, for more readability.

jaxley/modules/network.py
```diff
@@ -599,6 +631,12 @@ def _append_multiple_synapses(self, pre_nodes, post_nodes, synapse_type):
         # Define new synapses. Each row is one synapse.
         pre_nodes = pd.DataFrame({"pre_index": pre_nodes.index})
         post_nodes = pd.DataFrame({"post_index": post_nodes.index})
+
+        for key, value in synapse_type.node_params.items():
```
jnsbck, line 635: I think this could create some issues, if params are already defined in `ǹodes` via some mechanism and then a synapse is added that wants to use this same param i.e. in the presyn neuron. Might be a good idea to check if its already there

jaxley/synapses/alpha_synapse.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+from jax import Array
+
+from jaxley.solver_gate import exponential_euler
+from jaxley.synapses.synapse import Synapse
+
+
+class AlphaSynapse(Synapse):
+    r"""Alpha synapse which responds to binary pre-synaptic spike trains.
```
jnsbck, line 12: random idea, that might be non-sensical, but isnt binary a special case of [0,some_value]? In that case we could make it more general, then one could make use of spike amplitude.

PR #773, opened 2026-01-09
Fix dataclamp

kyralianaka:
As discussed with @michaeldeistler, I was unable to clamp multiple states (with different names) at once, only the last clamp was saved because the state names of the clamps weren't actually stored in data_clamp. The test I added actually fails on the current jaxley main.

The main fix is to now store the list of clamped state names in the data_clamp tuple where originally only the last state name was saved. This tuple is now (list of state names, list of state arrays, dataframe of indices). In the integration, state names and their state arrays are added to the externals dict separately in a loop.

### DIFF
--- a/tests/test_clamp.py
+++ b/tests/test_clamp.py
@@ -240,9 +240,13 @@ def test_multiple_data_clamps(SimpleNet):
     m1 = np.zeros_like(time) + 0.2
     m2 = np.zeros_like(time) + 0.3
 
-    data_clamps = net.cell(0).data_clamp("cell1_m", m1, None)
-    data_clamps = net.cell(1).data_clamp("cell2_m", m2, data_clamps)
-    soln = jx.integrate(net, delta_t=0.1, data_clamps=data_clamps)
+    def simulate(clamps):
+        data_clamps = net.cell(0).data_clamp("cell1_m", clamps[0], None)
+        data_clamps = net.cell(1).data_clamp("cell2_m", clamps[1], data_clamps)
+        return jx.integrate(net, delta_t=0.1, data_clamps=data_clamps)
+
+    jitted_simulate = jax.jit(simulate)
+    soln = jitted_simulate((m1, m2))
 
     net2 = SimpleNet(2, 1, 1)
     net2.cell(0).insert(HH(name="cell1"))


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 🐛 Bug fixes
+
+- Allow data_clamp to clamp multiple different states without silently only clamping the last state (#773, @kyralianaka)
```
jnsbck, line 10: is there an issue on this? If yes, could you link it to the PR and mention it here?

tests/test_clamp.py
```diff
@@ -225,3 +225,26 @@ def simulate(clamp):
 
     assert np.all(s[0, 1:] == -60.0)
     assert np.all(s[1, 1:] == -50.0)
+
+
+def test_multiple_data_clamps(SimpleNet):
```
jnsbck, line 230: you could also verify by matching a cell with 2 clamps to a cell with 2 data clanps

jnsbck: could be a better test case

jaxley/modules/base.py
```diff
@@ -2516,7 +2516,7 @@ def _external_input(
     def data_stimulate(
         self,
         current: ArrayLike,
-        data_stimuli: tuple[ArrayLike, pd.DataFrame] | None = None,
+        data_stimuli: tuple[List, List, pd.DataFrame] | None = None,
```
jnsbck: what would speak against a size (2,N) array? Aren't the lists the same size?

CHANGELOG.md
kyralianaka: there isn't an issue sorry, submitted the PR not long after I found the bug

jaxley/modules/base.py
kyralianaka: The list contains arrays of different sizes depending on how many cells were clamped for each of the state variables. For the current stimuli it doesn't matter because they all get concatenated in the end, but data_clamp and data_stimuli sharing the call to _data_external_input() forces them to have the same data structures.

I realized I need to fix some more of the type-hints though, will do that now :)

jnsbck: makes sense

tests/test_clamp.py
kyralianaka: comparing clamp and data_clamp might be a bit risky because a bug affecting one could affect the other, but I revised the test because it was more complicated than it needed to be. It basically does what you suggest, but just instead of integrating another clamped network, it just tests if the state vars are the correct clamp values

jnsbck: Since you're only testing for this specific case, ground truthing against clamp (which works and is well tested) I dont think that poses an issue.

jaxley/modules/base.py
```diff
@@ -2587,22 +2587,22 @@ def _data_external_input(
         ], "Number of comps and clamps do not match."
 
         if data_external_input is not None:
-            external_input = data_external_input[1]
-            external_input = jnp.concatenate([external_input, state_array])
-            inds = data_external_input[2]
+            external_state_names, external_input, inds = data_external_input
+            external_state_names.append(state_name)
+            external_input.append(state_array)
+            inds.append(view.index.to_numpy())
```
jnsbck, line 2593: Could this create issues with jit?

tests/test_clamp.py
```diff
@@ -225,3 +225,24 @@ def simulate(clamp):
...
+    net.cell(0).record("cell1_m")
+    net.cell(1).record("cell2_m")
+
+    time = np.arange(200, step=0.1)
+    m1 = np.zeros_like(time) + 0.2
+    m2 = np.zeros_like(time) + 0.3
+
+    data_clamps = net.cell(0).data_clamp("cell1_m", m1, None)
+    data_clamps = net.cell(1).data_clamp("cell2_m", m2, data_clamps)
+    soln = jx.integrate(net, delta_t=0.1, data_clamps=data_clamps)
```
jnsbck: Could you test the jitted case, since thats where data_clamp will most likely be used.

jaxley/modules/base.py
kyralianaka: I added jit to the test... In my experience it hasn't created any issues. 

Looking into the jax docs a _little_ bit it says "In JAX’s tracing and JIT compilation model, each element in a Python list or tuple is treated as a separate JAX variable, and individually processed and pushed to device," and "each entry of the list is handled as a separate input, resulting in a tracing & compilation overhead that grows linearly with the size of the list." The length of the list in this case should be the number of different variables you clamp, so I wouldn't expect performance issues in any normal use cases.

jnsbck: I was more worried about the `to_numpy`, but with it being tested its fine.

Issue #775, opened 2026-01-12
remove_observables from build_dynamic_state_utils doesn't work twice because of in-place updates on its argument


chaseking:
I ran into a problem where `remove_observables(all_states)` when called twice failed because it edits `all_states`. The problem is `remove_observables` calls `jaxley.utils.dynamics._remove_currents_from_states` which performs an in-place update on its argument.

It seems like the correct thing here is to make `_remove_currents_from_states` copy the states rather than editing them in-place with `del`. If `del` was used intentionally here, then please let me know.

chaseking:
Another reason in favor of copying, instead of del:
```
print(add_observables(remove_observables(all_states), all_params, delta_t).keys())
print(all_states.keys())

> dict_keys(['HH_h', 'HH_m', 'HH_n', 'v', 'i_HH'])
> dict_keys(['v', 'HH_m', 'HH_h', 'HH_n'])  # missing i_HH because all_states was edited
```
And additionally, the tree_maps used in the remove_observables method also effectively copy trees.

Matthijspals:
Hi!

my original motivation was assuming that we want to explicitly recompute currents anyway, e.g., like:


```python
# get dynamic states
dynamic_states = flatten(remove_observables(all_states))

# update dynamic states, e.g., update step of Kalman Filter
dynamic_states = dynamic_states * 1.1

# recompute currents based on new dynamic states and add them back to all_states
all_states = add_observables(unflatten(dynamic_states), all_params, delta_t=0.025)
# all states now contains I_HH again
```

However I agree that instead of using `del`, we can just return a _shallow_ copy of the dictionary instead, which doesn't really have a downside, yet has some benefits (keeping all_states keys in tact, consistency) as mentioned by @chaseking.

I guess all we need to do is:

change 
```python
    for key in current_keys:
        del states[key]
```
to 
```python
    return {k: v for k, v in states.items() if k not in current_keys}    
```

PR #776, opened 2026-01-12
Make _remove_currents_from_states copy states instead of editing in-place

chaseking:
Corrects in-place update by making copy of states instead (#775)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Add an identity transformation (`jaxley.optimize.transforms.IdentityTransform`) (#777, @chaseking)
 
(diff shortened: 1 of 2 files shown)


Matthijspals:
Hi, thanks for the pull request! See my comment #775, I agree we should do this!

Regarding the doc string
`    """Copies states without the currents through channels and synapses.
`
Maybe we can write
`    """Shallow copy of the states, excluding the currents through channels and synapses.
`
As we are not actually copying the contents in the dictionary, just pointers to them. (I think the `flatten` utility does make an explicit copy later on)

jnsbck (approved):

PR #777, opened 2026-01-12
Add identity transformation

chaseking:
It seemed clunky to have to use `AffineTransform(scale=1, shift=0)`, so this adds `IdentityTransform()`.

PR #779, opened 2026-01-12
graph with inhom compartments

NicolasRR:
Handles building a cell from a graph with inhomogeneous branches, i.e. different number of compartments.

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -12,12 +12,13 @@ jax.config.update("jax_platform_name", "cpu")
 
 os.environ["XLA_PYTHON_CLIENT_MEM_FRACTION"] = ".8"
 
+import numbers
+
 import jax.numpy as jnp
 import networkx as nx
 import numpy as np
 import pandas as pd
 import pytest
-import numbers
 
 import jaxley as jx
 from jaxley import connect
@@ -473,7 +474,7 @@ def test_from_graph_inhom_compartments(file):
         for key in a.keys():
             v1, v2 = a[key], b[key]
 
-            if isinstance(v1, numbers.Number) :
+            if isinstance(v1, numbers.Number):
                 if not np.isclose(v1, v2, atol=1e-5, equal_nan=True).all():
                     return False
             else:


jaxley/io/graph.py
```diff
@@ -1109,21 +1106,32 @@ def _build_module_scaffold(
...
             parents = (
                 default_parents if parent_branches is None else parent_branches[cell_id]
             )
+
+            branches = []
+            # size per branch, ordered by branch_index (stable / deterministic)
+            counts_by_branch = cell_groups.groupby("branch_index").size().sort_index()
+
+            for _, ncomps in counts_by_branch.items():
+                branches.append(Branch([comp for _ in range(int(ncomps))]))
```
jnsbck: Lets add a check to see if all ncomps are the same. If yes, then instantiate the branches as before, i.e. `[branch] * num_branches`. This is much faster since only one `Branch` instance needs to be created, rather than many.

jaxley/io/graph.py
```diff
@@ -1109,21 +1106,32 @@ def _build_module_scaffold(
...
     build_cache["compartment"] = [comp]
 
-    if return_type in return_types[1:]:
-        ncomps = idxs["branch_index"].value_counts().iloc[0]
-        branch = Branch([comp for _ in range(ncomps)])
-        build_cache["branch"] = [branch]
+    if return_type == return_types[1]:
+        branches = []
+        # size per branch, ordered by branch_index (stable / deterministic)
+        counts_by_branch = idxs.groupby("branch_index").size().sort_index()
```
jnsbck: Not sure if this is needed, since the if case only takes care of single branches, i.e. `len(counts_by_branch) == 1`

tests/test_graph.py
```diff
@@ -433,3 +433,36 @@ def test_trim_dendrites_of_swc():
...
+
+@pytest.mark.parametrize(
+    "file",
+    [
+        "morph_ca1_n120_single_point_soma.swc",
+        "morph_ca1_n120.swc",
+        "morph_l5pc_with_axon.swc",
+    ],
+)
+def test_from_graph_inhom_compartments(file):
```
jnsbck, line 446: I am not quite sure what this is testing. Did you mean to test if sth. like the following works?
```python
b1 = jx.Branch(2)
b2 = jx.Branch(3)
cell = jx.Cell([b1, b2], [-1,0])
cell_graph = to_graph(cell)

imported_cell = from_graph(cell_graph)
assert np.all(imported_cell.nodes["global_branch_index"] == cell.nodes["global_branch_index"])
```

jaxley/io/graph.py
NicolasRR: I restored the previous code for this block

tests/test_graph.py
NicolasRR: I changed the test to work with a cell with different number of compartments, please see new comment

tests/test_graph.py
NicolasRR, line 1: This is meant to test that we can use to_graph and from_graph successively and we get a cell with the identical nodes

jaxley/io/graph.py
```diff
@@ -1122,11 +1115,27 @@ def _build_module_scaffold(
...
-            cell = Cell(
-                [branch] * num_branches,
-                parents,
-                xyzr=xyzr[branch_counter : branch_counter + num_branches],
-            )
+
+            # size per branch, ordered by branch_index (stable / deterministic)
+            counts_by_branch = cell_groups.groupby("branch_index").size().sort_index()
+
+            inhom = np.std(counts_by_branch.values) > 1e-8
```
jnsbck: This is a bit nitpicky, but I think checking sth. like `np.all(counts_by_branch == counts_by_branch[0]))` would be more readable here.

tests/test_graph.py
jnsbck: Looks great! To reduce overhead you could also use one of the smaller morphologies, i.e. this one https://github.com/jaxleyverse/jaxley/blob/main/tests/swc_files/morph_ca1_n120_250.swc.

tests/test_graph.py
```diff
@@ -433,3 +433,50 @@ def test_trim_dendrites_of_swc():
...
+    # This cell has inhomogeneous compartments.
+    cell.initialize()
+    comp_graph = to_graph(cell, channels=True, synapses=True)
+
+    reconstructed_cell = from_graph(comp_graph)
+    reconstructed_graph = to_graph(reconstructed_cell, channels=True, synapses=True)
+
+    def node_match(a, b):
+        for key in a.keys():
+            try:
```
jnsbck: Lets use an `isinstance` rather than a try except if possible.

jnsbck: Also which attributes is the exception for?

In addition, you might wanna use `np.isclose` rather than `!=` on floats.

jnsbck (changes requested):
This is getting there! I left a few minor comments to be addressed. Also please rebase (#777 was just merged) and update the changelog. If all checks pass I am happy to merge this :)

tests/test_graph.py
NicolasRR: The exception is raised for xyzr which is an array
I also fixed the float comparison in the test

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ### 🧩 New features
 
 - Add an identity transformation (`jaxley.optimize.transforms.IdentityTransform`) (#777, @chaseking)
-
+- Add handling of inhomogeneous branches (#779,@NicolasRR)
```
jnsbck: ... for import and export of morphologies.

tests/test_graph.py
```diff
@@ -433,3 +433,54 @@ def test_trim_dendrites_of_swc():
...
+    comp_graph = to_graph(cell, channels=True, synapses=True)
+
+    reconstructed_cell = from_graph(comp_graph)
+    reconstructed_graph = to_graph(reconstructed_cell, channels=True, synapses=True)
+
+    def node_match(a, b):
+        for key in a.keys():
+            v1, v2 = a[key], b[key]
+
+            if isinstance(v1, float) or (
```
jnsbck: Maybe I missed something, but why would np.isclose(a,b).all() not work for everything, floats, ints, arrays?

NicolasRR: There are some graph element that are non numeric, i've update the if blocks to test whether data is numeric, see latest commit

jnsbck (approved):
Looks great! Thanks for iterating on this. Many thanks for contributing :)

Please make sure to run `isort` and `black`. If the checks pass, I can merge this.

PR #781, opened 2026-01-16
Add Troubleshooting Section to connect()

alexpejovic:
This is what I came up with regarding what you shortly mentioned in the meeting. Lmk if this needs changes!

### DIFF
--- a/jaxley/connect.py
+++ b/jaxley/connect.py
@@ -72,14 +72,23 @@ def connect(
     Troubleshooting
     ^^^^^^^^^^^^^^^
 
-    For large networks, using `connect()` might take a long time when selecting
+    For large networks, using ``connect()`` might take a long time when selecting
     a large amount of cells at once. When encountering this problem, one can
     connect the network using functions not in the public Jaxley API.
 
-    Example 3: Connect N/2 nodes with another set of N/2 nodes.
+    Below, we connect the first half of all compartments in a network to the
+    second half. Notice two things: First, we used the pandas function ``.loc`` on
+    ``.nodes`` directly, which is faster than the Jaxley method ``select`` (because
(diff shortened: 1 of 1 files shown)


jaxley/connect.py
```diff
@@ -68,6 +68,23 @@ def connect(
...
         print(net.edges)
+
+    Troubleshooting
+    ^^^^^^^^^^^^^^^
+
+    For large networks, using `connect()` might take a long time when selecting
+    a large amount of cells at once. When encountering this problem, one can
+    connect the network using functions not in the public Jaxley API.
+
+    Example 3: Connect N/2 nodes with another set of N/2 nodes.
```
michaeldeistler: Please remove `Example 3:`. Also, don't use the term "nodes", just use "compartments".

I suggest:

Below, we connect the first half of all compartments in a network to the second half. Notice two things: First, we used the pandas function `.loc` on `.nodes` directly, which is faster than the `Jaxley` method `select` (because `select` builds a `View` of the `net`). Second, we use `_append_multiple_synapses()`, which is faster than `connect()`.

[If you know the reason why `_append_multiple_synapses` is faster than `connect`, please feel free to add it in brackets :) ]

jaxley/connect.py
```diff
@@ -68,6 +68,23 @@ def connect(
...
+
+    ::
```
michaeldeistler: Let's add
```python
cell = jx.Cell()
net = jx.Network([cell for _ in range(N)])
```

michaeldeistler (approved):
Added two comments below. Good to go once addressed. Thanks!

jaxley/connect.py
alexpejovic: `_append_multiple_synapses` isn't faster than `connect`, `select` is the only thing making things slower.

PR #803, opened 2026-06-14
enh: add uniform distribution to utils (#707)

### LINKED ISSUE #707: Add optimization utils / voltage trace features
Add `jaxley.optimize.features` / `jaxley.optimize.distributions` or just both of these things to `optimize.utils`

Here would be an example of this:

```python
# Multidimensional uniform distribution in jax
class Uniform:
    """Uniform distribution with sample and log_prob methods."""
    def __init__(self, lower: float, upper: float) -> None:
        self.lower = lower
        self.upper = upper

    def sample(self, key: jnp.ndarray, shape: Tuple[int, ...] = (1,)) -> jnp.ndarray:
        """Samples from the uniform distribution.

        Args:
            key: A JAX random key.
            shape: Sample shape.

        Returns:
            Samples from the uniform distribution.
        """
        return jax.random.uniform(key, shape=shape, minval=self.lower, maxval=self.upper)

    def log_prob(self, x: jnp.ndarray) -> jnp.ndarray:
        """Computes the log probability of the uniform distribution.

        Args:
            x: The input to compute the log probability for.

        Returns:
            The log probability of the uniform distribution.
        """
        in_bounds = (x >= self.lower) & (x <= self.upper)
        return jnp.where(in_bounds, -jnp.log(self.upper - self.lower), -jnp.inf)

class ProductDistribution:
    """Product distribution of multiple distributions; p(x,y,z) = p(x)p(y)p(z)."""
    def __init__(self, dists: Dict[str, Uniform]) -> None:
        self.dists = dists

    def sample(self, key: jnp.ndarray, shape: Tuple[int, ...] = (1,)) -> Dict[str, jnp.ndarray]:
        split_keys = jax.random.split(key, len(self.dists))
        split_keys = {k: v for k, v in zip(self.dists.keys(), split_keys)}
        return jax.tree_util.tree_map(lambda k, d: d.sample(k, shape), split_keys, self.dists)

    def log_prob(self, x: Dict[str, jnp.ndarray]) -> jnp.ndarray:
        log_probs = jax.tree_util.tree_map(lambda x, d: d.log_prob(x), x, self.dists)
        return jnp.sum(jnp.array(jax.tree_leaves(log_probs)), axis=0)

class BoxUniform(ProductDistribution):
    """Multi-dimensional uniform distribution.
    
    Args:
        bounds: A dictionary of parameter names and their bounds.
    """
    def __init__(self, bounds: Dict[str, Tuple[float, float]]) -> None:
        dists = {k: Uniform(*v) for k, v in bounds.items()}
        super().__init__(dists)
```

```python
prior = BoxUniform(bounds)
key = jax.random.PRNGKey(23)
obs_params = prior.sample(key, (1,))
```

iamkzntsv:
Partially fixes #707.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Implements Brette et al. (2005), 'Adaptive exponential integrate-and-fire model 
 - Add an identity transformation (`jaxley.optimize.transforms.IdentityTransform`) (#777, @chaseking)
 - Add handling of inhomogeneous branches for import and export of morphologies. (#779,@NicolasRR)
 - Add an logistic transformation (`jaxley.optimize.transforms.LogisticTransform`) (#788, @jnsbck)
+- Add a uniform distribution (`jaxley.optimize.utils.Uniform`) (#707, @iamkzntsv)
 
 ### 🐛 Bug fixes
 
--- a/jaxley/optimize/utils.py
+++ b/jaxley/optimize/utils.py
@@ -15,7 +15,8 @@ def l2_norm(x: "PyTree") -> jnp.array:
 
 
 class Uniform:
-    """Multidimensional uniform distribution with sample and logprob methods."""
+    """Uniform distribution with sample and log_prob methods."""
+
     def __init__(self, lower: float, upper: float):
         self.lower = lower
         self.upper = upper
@@ -30,7 +31,9 @@ class Uniform:
         Returns:
             Samples from the uniform distribution.
         """
-        return jax.random.uniform(key, shape, minval=self.lower, maxval=self.upper)
+        return jax.random.uniform(
+            key, shape=shape, minval=self.lower, maxval=self.upper
+        )
 
     def log_prob(self, x: jnp.ndarray) -> jnp.ndarray:
         """Computes the log probability of the uniform distribution.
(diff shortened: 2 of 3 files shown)


michaeldeistler:
Hi @iamkzntsv, thanks a lot for getting this started! Once finished, please request reviews of me and @jnsbck. Looking forward!

iamkzntsv:
Hi @michaeldeistler and @jnsbck! I think the only way I can request a review is by tagging you in the comment.

I've only added the `Uniform` distribution from #707 to keep this first PR lightweight, and I would be very happy to implement the remaining distributions in a follow-up PR :) Please let me know if you need any tweaks here.

Sergey

michaeldeistler (changes requested):
Thanks a lot!

Please add a few tests in the `tests` folder. In particular, please test:
- the shape of samples (dimension (N, D), for N samples of dimension D)
- the shape of log-probs (which should be (N,) for N samples, each of dimension D)

In addition, I am confused about the scope of this distribution: The docstring says "Multidimensional", but the datatype of `lower` and `upper` is specified as `float`. Which is it?

michaeldeistler:
Please also add a line to the `CHANGELOG.md` :)

iamkzntsv:
Hi @michaeldeistler and @jnsbck! 

Thanks for your feedback. You're right, "Multidimensional" in the docstring indeed looked ambiguous, so I removed it. I have also added the reduction over non-batch dimensions to `Uniform.log_prob()` and added a few tests. Thank you :)

Sergey

PR #809, opened 2026-09-11
Fix jitting solvers with constant matrices broken under jit(vmap) on certain JAX versions

Matthijspals:
Between **jax/jaxlib ≈ 0.5.3 and 0.8.1**, the pattern below (used in the Jaxley solver) seems broken when jitting a `vmap` over several columns of a matrix that is a constant inside the function (i.e., not passed in as an argument):

```python
solves.at[i].set(voltages[i] + delta_t * constant_terms[i])
```

`jit(f)(B)` is fine but `jit(lambda: f(B))()` returns wrong values. See [jax-ml/jax#33479](https://github.com/jax-ml/jax/issues/33479) / [openxla/xla#34260](https://github.com/openxla/xla/issues/34260) (fixed in jaxlib ≥ 0.8.2).

**Minimal example:**
```python
import jax, jax.numpy as jnp

B = jax.random.normal(jax.random.PRNGKey(0), (10, 8))
idx = jnp.array([1, 3, 5, 7])

def col_fn(rhs):
    v = jnp.zeros(10).at[:10].set(rhs)
    c = jnp.zeros(10)
    return jnp.zeros(10).at[idx].set(v[idx] + 0.025 * c[idx])

vmap_fn = lambda M: jax.vmap(col_fn, in_axes=1, out_axes=1)(M)
print(jnp.max(jnp.abs(jax.jit(vmap_fn)(B) - jax.jit(lambda: vmap_fn(B))())))
```

This affects Jaxley’s implicit voltage solves (DHS, sparse, Stone) when applying them to each column of a fixed matrix `B` inside a jitted function.

**Fix:** 
write it as:
```python
solves = solves.at[i].set(voltages[i])
solves = solves.at[i].add((delta_t * constant_terms)[i])
```

**Added a test** that `jit(vmap)(B)` matches `jit(lambda: vmap(B))()`

Disclaimer: diagnosed and fixed with help of cursor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Implements Brette et al. (2005), 'Adaptive exponential integrate-and-fire model 
 - Fix single point branch plotting with type="comp" (#797, @jnsbck)
 - jx.integrate took O(n^2) time with n compartments on the backwards pass. Instead of backpropagating through the forward solve, we now use a custom_jvp (another tridiagonal solve, which is O(n)) (#795 @manuelgloeckler)
 - Fix compatibility with newer JAX versions by avoiding the deprecated `jnp.clip(..., a_max=...)` keyword in `solver_gate.save_exp` (#798, @lorenzosquadrani)
+- Fix incorrect implicit voltage RHS under `jit(vmap)` on jax/jaxlib 0.5.3–0.8.1 by assembling as set-then-add instead of `v[i] + dt * c[i]` (@matthijspals)
 
 # 0.13.0
 
--- a/jaxley/solver_voltage.py
+++ b/jaxley/solver_voltage.py
@@ -186,10 +186,12 @@ def step_voltage_implicit_with_dhs_solve(
     if len(sinks) > 0:
         diags = diags.at[sinks].add(axial_conductances)
 
-    # Build solve.
+    # Build solve. Use set + add (not `v[i] + dt * c[i]`) to avoid an XLA
+    # constant-folding bug under jit(vmap) on jax 0.5.3–0.8.1 (jax#33479).
     solves = jnp.zeros(n_nodes)
-    solves = solves.at[internal_node_inds].set(
-        voltages[internal_node_inds] + delta_t * constant_terms[internal_node_inds]
+    solves = solves.at[internal_node_inds].set(voltages[internal_node_inds])
+    solves = solves.at[internal_node_inds].add(
+        (delta_t * constant_terms)[internal_node_inds]
     )
 
     # Why `n_nodes > 1`? For compartments (or point neurons), we save computation and
@@ -377,8 +379,9 @@ def step_voltage_implicit_with_jax_spsolve(
 
     # Build solve.
     solves = jnp.zeros(n_nodes)
-    solves = solves.at[internal_node_inds].set(
-        voltages[internal_node_inds] + delta_t * constant_terms[internal_node_inds]
+    solves = solves.at[internal_node_inds].set(voltages[internal_node_inds])
+    solves = solves.at[internal_node_inds].add(
+        (delta_t * constant_terms)[internal_node_inds]
     )
 
     # Concatenate diagonals and off-diagonals (which are just `-axial_conductances`).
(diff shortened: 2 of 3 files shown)


jnsbck:
Hey Matthijs, 

thanks for going through all the hassle to track this down and for submitting this! This looks like a super curious bug! Am I getting this correctly that the fix basically patches jaxley for jaxlib: v0.5.3 - v0.8.1, but for newer version it does not matter or the new pattern also just plain better "code"? Did you check if this impacts performance on the newer jax versions? If yes, I'd be hesitant to merge it into main and rather list it as a patch for the affected versions! 

Best,

Jonas

Matthijspals:
Hi, thanks for getting back to me! Yes, this affects only those jax versions, and indeed I think my proposed solution was actually slower (1.5–2× slower in isolation, but probably in a full solve not so different). In any case, I think there is another way to avoid this error that seems to come without performance cost (as long as *i* / *internal_nodes* runs along is most entries)
```python
# broken for the operation above: 
solves.at[i].set(voltages[i] + delta_t * constant_terms[i])

# okay:
solves.at[i].set((voltages + delta_t * constant_terms)[i])
```
To quickly verify this, you can ran this example
```python
import jax, jax.numpy as jnp

B = jax.random.normal(jax.random.PRNGKey(0), (10, 8))
idx = jnp.array([1, 3, 5, 7])

def col_fn(rhs):
    v = jnp.zeros(10).at[:10].set(rhs)
    c = jnp.zeros(10)
    return jnp.zeros(10).at[idx].set(v[idx] + 0.025 * c[idx])

def col_fn2(rhs):
    v = jnp.zeros(10).at[:10].set(rhs)
    c = jnp.zeros(10)
    return jnp.zeros(10).at[idx].set((v + 0.025 * c)[idx])

vmap_fn = lambda M: jax.vmap(col_fn, in_axes=1, out_axes=1)(M)
print(jnp.max(jnp.abs(jax.jit(vmap_fn)(B) - jax.jit(lambda: vmap_fn(B))())))

vmap_fn2 = lambda M: jax.vmap(col_fn2, in_axes=1, out_axes=1)(M)
print(jnp.max(jnp.abs(jax.jit(vmap_fn2)(B) - jax.jit(lambda: vmap_fn2(B))())))
```

On Jax 0.6.2 the first print is 2.71, the second is 0.

I'll update the pull request.

jnsbck:
Cool! This looks much simpler. I guess this is only bad for highly branched morphologies (as you also point out I think). Since the issue seems limited to older jax versions, I would propose to only merge the test, document this and release the patch just on github. Let me know if that works. :)

Also @michaeldeistler feel free to give a thumbs up if you happen to look at your email :)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Implements Brette et al. (2005), 'Adaptive exponential integrate-and-fire model
 - Fix single point branch plotting with type="comp" (#797, @jnsbck)
 - jx.integrate took O(n^2) time with n compartments on the backwards pass. Instead of backpropagating through the forward solve, we now use a custom_jvp (another tridiagonal solve, which is O(n)) (#795 @manuelgloeckler)
 - Fix compatibility with newer JAX versions by avoiding the deprecated `jnp.clip(..., a_max=...)` keyword in `solver_gate.save_exp` (#798, @lorenzosquadrani)
+- Fix edge case incorrect implicit voltage RHS under `jit(vmap)` on jax/jaxlib 0.5.3–0.8.1 by assembling as `(v + dt * c)[i]` instead of `v[i] + dt * c[i]` (@matthijspals)
```
michaeldeistler, line 19: 0.14.0 is already released, could you add it under a new 0.15.0?

michaeldeistler (approved):
Awesome, thanks a ton Matthijs!

CHANGELOG.md
jnsbck: I'd advocate for 0.14.1, since its only a patch.

jnsbck:
Another option would be to make the behaviour depend on jax._\_version__, but this might be overkill.

